PR #4, opened 2020-06-19
Minor fixes to the sources and requirements

Tiboris:
Fixing:
db_driver.load() to take no params as the path is handled in init
ctx object should have ctx.obj to be accessed with key
Adding:
requirements were missing when installing in clear environment
yaml lib helper functions

### DIFF
--- a/src/aiohabit/run.py
+++ b/src/aiohabit/run.py
@@ -18,7 +18,7 @@ import asyncio
 import click
 
 from .dbdrivers.file import FileDBDriver
-from .utils import load_json, print_obj
+from .utils import load_yaml, print_obj
 from .config import ProvisioningConfig
 from .actions.destroy import Destroy
 from .actions.up import Up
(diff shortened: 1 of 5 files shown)


src/aiohabit/run.py
```diff
@@ -63,8 +63,8 @@ def aiohabitcli(ctx, config, db):
     """Multihost human friendly provisioner."""
     init_providers()
     ctx.ensure_object(dict)
-    ctx[DB] = init_db(db)
-    ctx[CONFIG] = init_config(config)
+    setattr(ctx, DB, init_db(db))
```
pvoborni, line 66: In other project I've fixed this issue by:

```python
@click.pass_context
def managercli(ctx, config, verbose):
    """Uploads new OpenStack images from OS composes"""
    ctx.ensure_object(dict)
    ctx.obj[CONFIG] = init_config(config)
```

Tiboris: So what what would be our preference?

pvoborni: The method which I wrote. It is also used in the official documentation: https://click.palletsprojects.com/en/7.x/commands/#nested-handling-and-contexts

pvoborni: I.e. my initial implementation in aiohabit simply forgot the `obj` part.

Tiboris: Let me update it.

PR #7, opened 2020-07-14
Ansible Inventory and pytest-multihost output modules

pvoborni:
## Add Ansible Inventory and pytest-multihost output modules

And enhance "up" action to use them. Add also "output" action which  calls them separately. It can be used for recreation of deleted ones  or testing.
    
## hosts: adding missing properties
    
So that they can be work with in output modules.

### DIFF
--- a/src/aiohabit/outputs/ansible_inventory.py
+++ b/src/aiohabit/outputs/ansible_inventory.py
@@ -1,0 +1,143 @@
+# Copyright 2020 Red Hat Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Ansible inventory output module."""
+
+from copy import deepcopy
+from aiohabit.outputs.utils import is_windows_host, resolve_hostname
+from aiohabit.utils import get_config_value, get_host_from_metadata, save_yaml
+
+DEFAULT_INVENTORY_PATH = "aiohabit-inventory.yaml"
+DEFAULT_INVENTORY_LAYOUT = {"all": {"children": {}, "hosts": {}}}
+
+
+def copy_meta_attrs(host, meta_host, attrs):
+    """Copy attributes from host metadata entry and prefixing them with meta."""
+    for attr in attrs:
+        val = meta_host.get(attr)
+        if val:
+            host[f"meta_{attr}"] = val
+
+
+def find_group(inventory, group_names, top=True):
+    """Find first matching group in inventory which has one of the provided names."""
+    if not inventory:
+        return None
+
+    groups = inventory.keys()
+    for group in groups:
+        if group in group_names:
+            return inventory[group]
+
+    for group in groups:
+        g = inventory[group]
+
+        if "children" not in g:
+            continue
+
+        found = find_group(g["children"], group_names, top=False)
+        if found:
+            return found
+
+    if top:
+        return inventory["all"]
+
+    return None
+
+
+class AnsibleInventoryOutput:
+    """
+    Generate Ansible inventory with provisioned machines.
+
+    The resulting inventory combines data from provisioning config, job metadata
+    files and information in DB.
+    """
+
+    def __init__(self, config, db, metadata, path=DEFAULT_INVENTORY_PATH):
+        """Init the output module."""
+        self._config = config
+        self._db = db
+        self._metadata = metadata
+        self._path = path
+
+    def create_ansible_host(self, name):
+        """Create host entry for Ansible inventory."""
+        meta_host, meta_domain = get_host_from_metadata(self._metadata, name)
+        db_host = self._db.hosts[name]
+
+        ip = db_host.ips[0]
+        ansible_host = resolve_hostname(ip) or ip
+
+        python = (
+            self._config["python"][meta_host["os"]] or self._config["python"]["default"]
+        )
+
+        default_user = get_config_value(self._config["users"], meta_host["os"])
+        ansible_user = db_host.username or meta_host.get("username", default_user)
+
+        # Common attributes
+        host_info = {
+            "ansible_host": ansible_host,
+            "ansible_python_interpreter": python,
+            "ansible_user": ansible_user,
+            "meta_fqdn": name,
+            "meta_domain": meta_domain["name"],
+            "meta_provider_id": db_host.id,
+            "meta_ip": ip,
+            "meta_dc_record": ",".join(
+                "DC=%s" % dc for dc in meta_domain["name"].split(".")
+            ),
+        }
+        copy_meta_attrs(host_info, meta_host, ["os", "role", "netbios"])
+
+        if "parent" in meta_domain:
+            host_info["parent_domain"] = meta_domain["parent"]
+
+        if is_windows_host(meta_host):
+            password = meta_host.get(
+                "password", self._config["mhcfg"]["ad_admin_password"]
+            )
+            host_info.update(
+                {
+                    "ansible_password": password,
+                    "ansible_port": 5986,
+                    "ansible_connection": "winrm",
+                    "ansible_winrm_server_cert_validation": "ignore",
+                    "meta_domain_level": meta_host.get("domain_level", "top"),
+                }
+            )
+        return host_info
+
+    def create_inventory(self):
+        """Create the Ansible inventory in dict form."""
+        provisioned = self._db.hosts
+        inventory = deepcopy(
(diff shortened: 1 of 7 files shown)


pvoborni:
@Tiboris or @netoarmando , let me know if I should merge this.

src/aiohabit/run.py
```diff
@@ -24,6 +24,7 @@
 from aiohabit.config import ProvisioningConfig
 from aiohabit.actions.destroy import Destroy
 from aiohabit.actions.up import Up
+from aiohabit.actions.output import Output
```
netoarmando, line 27: ```ModuleNotFoundError: No module named 'aiohabit.actions.output'```

I think that if this won't be implemented with this PR you can remove the changes `run.py` and save them for later.

src/aiohabit/outputs/utils.py
```diff
@@ -0,0 +1,48 @@
...
+
+def resolve_hostname(ip):
+    """Resolve IP address to hostname."""
+    try:
+        return socket.gethostbyaddr(ip)[0]
+    except socket_error:
+        return None
+
+
+def get_val(values):
```
netoarmando, line 29: This is called mostly passing a list of only 2 items, so could be replaced with: ```result = first or second```.
But, if you think this is valuable I would rename it to something like `get_first_valid`.

src/aiohabit/outputs/pytest_multihost.py
```diff
@@ -0,0 +1,128 @@
...
+
+        # ssh_key_filename must be absolute as it can be used from a different
+        # working directory, e.g. running tests from git repo
+        ssh_key_filename = mhcfg.get("ssh_key_filename")
+        if ssh_key_filename and not ssh_key_filename.startswith("~"):
+            mhcfg["ssh_key_filename"] = os.path.abspath(ssh_key_filename)
+
+        return mhcfg
+
+    def create_output(self):
```
netoarmando, line 125: I guess this will be the only common method used by all "output classes", right?

src/aiohabit/run.py
pvoborni: I forgot to  `git add` the file. Corrected.

src/aiohabit/outputs/utils.py
pvoborni: Renamed. I mainly implemented it for the case with 3 items. But after the rename and push, I see that it is used a lot for the 2 value cases where it indeed can use the 'or'.

src/aiohabit/outputs/pytest_multihost.py
pvoborni: That was the intend, but we don't need to stick with it if it will not make sense.

src/aiohabit/outputs/utils.py
pvoborni: So after second, more careful read. I've rewrote it as suggested to the 'or' form and in the end removed the method completely. And the code looks better.

pvoborni:
Fixed also comment in Ansible Inventory as it was incorrectly referencing to Linchpin.

src/aiohabit/outputs/ansible_inventory.py
```diff
@@ -0,0 +1,141 @@
...
+
+    def create_ansible_host(self, name):
+        """Create host entry for Ansible inventory."""
+        meta_host, meta_domain = get_host_from_metadata(self._metadata, name)
+        db_host = self._db.hosts[name]
+
+        ip = db_host.ips[0]
+        ansible_host = resolve_hostname(ip) or ip
+
+        python = self._config["python"] or meta_host["os"]
```
Tiboris, line 82: First of all, thanks for the work done!
When trying the generation i bumped into this line in utput:
```yaml
ansible_python_interpreter: *id001
```
which may be because the 
```python
self._config["python"] is a dictionary
```
it is probably some minor bug or the difference in provisioning config which i use:
```yaml
# ansible_python_interpreter
python:
    fedora-28: /usr/bin/python3
    fedora-29: /usr/bin/python3
    fedora-30: /usr/bin/python3
    fedora-31: /usr/bin/python3
    fedora-32: /usr/bin/python3
    fedora-rawhide: /usr/bin/python3
    rhel-6.10: /usr/bin/python
    rhel-7.0: /usr/bin/python
    rhel-7.1: /usr/bin/python
    rhel-7.2: /usr/bin/python
    rhel-7.3: /usr/bin/python
    rhel-7.4: /usr/bin/python
    rhel-7.5: /usr/bin/python
    rhel-7.5-atomic: /usr/bin/python
    rhel-7.6-atomic: /usr/bin/python
    default: /usr/libexec/platform-python
```

Tiboris: This worked for me:
```suggestion
        python = self._config["python"][meta_host["os"]] or self._config["python"]["default"]
```
could you please check if that match the intention?

result
```yaml
      ansible_python_interpreter: /usr/bin/python3
```

Tiboris:
Looks all right now from my side. @netoarmando can we merge this?

src/aiohabit/utils.py
```diff
@@ -133,6 +133,18 @@ def print_obj(obj):
     print(object2json(obj))
 
 
+def get_host_from_metadata(metadata, name):
```
netoarmando, line 136: `name` is not used, this function always returns the first host, is it expected?

src/aiohabit/run.py
```diff
@@ -113,6 +118,18 @@ async def destroy(ctx, metadata):
     await destroy_action.destroy()
 
 
+@aiohabitcli.command()
+@click.pass_context
+@click.argument("metadata")
+@async_run
+async def output(ctx, metadata):
+    """Destroy provisioned hosts."""
```
netoarmando, line 126: Docstring must be updated.

src/aiohabit/outputs/ansible_inventory.py
```diff
@@ -0,0 +1,143 @@
...
+def copy_meta_attrs(host, meta_host, attrs):
+    """Copy attributes from host metadata entry and prefixing them with meta."""
+    for attr in attrs:
+        val = meta_host.get(attr)
+        if val:
+            host[f"meta_{attr}"] = val
+
+
+def find_group(inventory, group_names, top=True):
+    """Find first matching group in inventory which has one of the provided names."""
```
netoarmando, line 34: If none of the names in `group_names` is found at the inventory and `top` is `True` then all groups will be returned, did I get it right?

pvoborni: I'd say that "all" group will be returned. It works with group layout defined in, e.g.: 
```
DEFAULT_INVENTORY_LAYOUT = {"all": {"children": [], "hosts": {}}}
```

But it can be more complex (e.g. for testing FreeIPA) (now defined in provisioning config) where the original code was like:
```
children = {
        'ipa': {
            'children': {
                'ipaserver': {},
                'ipaclient': {},

            }
        },
        'ad': {
            'children': {
                'ad_root': {},
                'ad_subdomain': {},
                'ad_treedomain': {},
                'ad_member': {}
            }
        },
        'linux': {
            'children': {
                'ipa': {},
                'runner': {}
            }
        },
        'windows': {
            'children': {
                'ad': {}
            }
        },

    }
    inventory = {
        'all': {
            'children': children,
            'hosts': {},
        }
    }
```

src/aiohabit/utils.py
pvoborni: good catch, fixed, filtering by name added

src/aiohabit/run.py
pvoborni: fixed

src/aiohabit/utils.py
```diff
@@ -133,6 +133,23 @@ def print_obj(obj):
...
+    Returns:
+    (host, domain)
+    """
+    domains = metadata.get("domains", [])
+    match = [
+        (host, domain)
+        for domain in domains
+        for host in domain.get("hosts", [])
+        if host["name"] == name
+    ]
```
netoarmando, line 149: Final comment: I think we should avoid list comprehensions in complex cases like that, this is much easier to read and debug:
```python
for domain in domains:
    for host in domain.get("hosts", []): 
        if host["name"] == name:
            return host, domain
````
It even makes clear that if no host with `name` is found then only 1 variable will be returned ("`None`" instead of "`None, None`"), but we don't need to handle this right now and it will fail explicitly in the future.

Comment (pvoborni) [reply src/aiohabit/utils.py]: You are right, this code is better and more readable. Changed.

**Comment by pvoborni** [comment]
I've also found an issue that the resulting Ansible inventory was not correct. PR updated with fix:
```diff
--- a/src/aiohabit/outputs/ansible_inventory.py
+++ b/src/aiohabit/outputs/ansible_inventory.py
@@ -19,7 +19,7 @@ from aiohabit.outputs.utils import is_windows_host, resolve_hostname
 from aiohabit.utils import get_config_value, get_host_from_metadata, save_yaml
 
 DEFAULT_INVENTORY_PATH = "aiohabit-inventory.yaml"
-DEFAULT_INVENTORY_LAYOUT = {"all": {"children": [], "hosts": {}}}
+DEFAULT_INVENTORY_LAYOUT = {"all": {"children": {}, "hosts": {}}}
```

now e.g. `ansible -i aiohabit-inventory.yaml -m shell -a "hostname" all` should work

PR #8, opened 2020-07-18
Add "list" and  "ssh" actions

pvoborni:
This is a follow up after(based on) https://github.com/pvoborni/aiohabit/pull/7 

## List action
To quickly show content of DB in human readable form

## SSH action
To make it very easy to SSH into a host based on value in various  configuration objects.
    
It is usually very time-consuming to "cat" e.g. inventory, copy the real hostname/IP, write all the SSH options including determining if it should use SSH key and what is its location.
    
The interactive mode even saves time with writing the fake hostname.

The bad UX of these actions is a need to provide metadata/config paths. In some other PR, we should figure out some reasonable default so that ppl can simply execute: `aiohabit list` and `aiohabit ssh`

### DIFF
--- a/src/aiohabit/actions/list.py
+++ b/src/aiohabit/actions/list.py
@@ -1,0 +1,33 @@
+# Copyright 2020 Red Hat Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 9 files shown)


netoarmando:
@pvoborniEverything looks good, but I have two considerations (which I want to fill bugs for):

1. There's a bug in `AWSProvider.to_host` method that is breaking the SSH action: https://github.com/pvoborni/aiohabit/blob/97f482e2e9338c618fdec42bb2f4970057725976/src/aiohabit/providers/aws.py#L216
That should line must be replaced by:
```python
STATUS_MAP[host_info.get("status")],
```
This could be done in this PR, but we also need to better map the statuses among providers.

2. We could improve the names of the classes to reflect what they do, like rename all action classes to `SomethingAction`, e.g: `SSHAction`, `ListAction`.

---

Also, I couldn't ssh into AWS hosts, I tried to force the private key without success. Since I don't think that's related to the changes you are providing, I'm fine with all here.

pvoborni:
Thanks Armando for the review. I agree with both points.

PR #9, opened 2020-07-24
Add basic logging mechanism

netoarmando:
Configure a logger to be used in the whole application.

---

Based on the table [here](https://docs.python.org/3/howto/logging.html#when-to-use-logging, I wonder if we should keep all the prints for stdout and stderr messages instead of replacing them.

### DIFF
--- a/src/aiohabit/providers/openstack.py
+++ b/src/aiohabit/providers/openstack.py
@@ -15,6 +15,7 @@
 """OpenStack provider."""
 
 import asyncio
+import logging
 from copy import deepcopy
 from datetime import datetime, timedelta
 from asyncopenstackclient import GlanceClient, AuthPassword
@@ -31,6 +32,8 @@ from aiohabit.host import (
     STATUS_OTHER,
 )
 from aiohabit.errors import ServerNotFoundError, ValidationError, ProvisioningError
+
+logger = logging.getLogger(__name__)
 
 # Docs
 # https://github.com/DreamLab/AsyncOpenStackClient
@@ -85,6 +88,7 @@ class OpenStackProvider(Provider):
         * account limits (max and current usage of vCPUs, memory, ...)
         """
         # session expects that credentials will be set via env variables
+        logger.info("Initializing OpenStack provider")
         self.session = AuthPassword()
         self.nova = ExtraNovaClient(session=self.session)
         self.glance = GlanceClient(session=self.session)
@@ -96,7 +100,7 @@ class OpenStackProvider(Provider):
         )
         login_end = datetime.now()
         login_duration = login_end - login_start
-        print(f"Login duration {login_duration}")
+        logger.info(f"Login duration {login_duration}")
 
         object_start = datetime.now()
         flavors, images, limits, networks, ips = await asyncio.gather(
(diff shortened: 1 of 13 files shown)


pvoborni:
It will probably depend on the action to run.  Actions like 'up' and 'destroy' should probably log almost everything. They will be used in CIs so the log should be available. I think that we can log more info (e.g. info from providers) to DEBUG level than we now print to console. 

For human 'console' actions like 'list' and 'ssh' we probably will mostly log to INFO and print.

netoarmando:
> It will probably depend on the action to run. Actions like 'up' and 'destroy' should probably log almost everything. They will be used in CIs so the log should be available. I think that we can log more info (e.g. info from providers) to DEBUG level than we now print to console.
> 
> For human 'console' actions like 'list' and 'ssh' we probably will mostly log to INFO and print.

Agreed! Let me enhance this.

pvoborni:
I see that there is used pattern:
```
        logger.info(f"Hosts to delete: {names}")
        print(f"Hosts to delete: {names}")
```
Do we want this code duplicity with strings?  

In the previous comment I thought that we will:
* log INFO level into console and DEBUG level into file implicitly. 
* log DEBUG into console when --debug is passed
* have separate `print()` statements only for text which we don't want to log into file.

I'm also open to not have the console logger but in such case we might want to have some utility function like
```
def print_log(logger, msg, level="INFO"):    
    file = sys.stdout
    if level in ("ERROR", "CRITICAL"):
       file = syst.stderr
    print(msg, file=file)
    logger.log(level, msg)
```

netoarmando:
@pvoborni it was hard to feel comfortable with the idea of logging+printing a CLI application. Specially because the common practice is to log to stdout. But I think your points now made it clear to me. 

About the duplication, I wonder if most of the strings would be the same or not. However, I guess all of this is fixed by having a custom `print` function. It looks strange, but I think is the best solution.

netoarmando:
Earlier today @pvoborni and I talked about this and we agreed to the following points:
- Remove all `print` and replace them with `logger.info`
- One handler would output the messages to the console without timestamps
- Another handler will display more info, such as the timestamps

src/aiohabit/providers/openstack.py
```diff
@@ -109,7 +113,7 @@ async def init(self, image_names=None):
         self.limits = limits
         object_end = datetime.now()
         object_duration = object_end - object_start
-        print(f"Open Stack environment objects " f"load duration: {object_duration}")
+        logger.info(f"Open Stack environment objects " f"load duration: {object_duration}")
```
Tiboris, line 116: Line is too long

src/aiohabit/providers/openstack.py
```diff
@@ -312,8 +316,8 @@ async def can_provision(self, reqs):
         req_vcpus = used_vcpus + vcpus
         req_memory = used_memory + ram
 
-        print(f"Required vcpus: {vcpus}, " f"used: {used_vcpus}, max: {limit_vcpus}")
-        print(f"Required ram: {ram}, used: {used_memory}, max: {limit_memory}")
+        logger.info(f"Required vcpus: {vcpus}, " f"used: {used_vcpus}, max: {limit_vcpus}")
```
Tiboris, line 319: Line is too long

Tiboris:
@pvoborni  Let us merge this as well. Thanks @netoarmando

PR #11, opened 2020-07-26
AWS host: Return the right status

netoarmando:
Fixing a small issue, `STATUS_MAP` was defined but not used. 

Issue described at https://github.com/pvoborni/aiohabit/pull/8#issuecomment-663308754.

### DIFF
--- a/src/aiohabit/providers/aws.py
+++ b/src/aiohabit/providers/aws.py
@@ -25,6 +25,7 @@ from aiohabit.host import (
     STATUS_PROVISIONING,
     STATUS_DELETED,
     STATUS_ERROR,
+    STATUS_OTHER,
 )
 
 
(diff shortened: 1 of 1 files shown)


src/aiohabit/providers/aws.py
```diff
@@ -213,7 +213,7 @@ def to_host(self, provisioning_result):
             host_info.get("id"),
             host_info.get("name"),
             host_info.get("addresses"),
-            host_info.get("status"),
+            STATUS_MAP[host_info.get("status")],
```
pvoborni, line 216: The STATUS_MAP doesn't cover all statuses. So we might want something similar as in OpenStack provider where is  `status = STATUS_MAP.get(provisioning_result.get("status"), STATUS_OTHER)`

Tiboris:
@pvoborni  let us merge this! thanks for the fix!

PR #14, opened 2020-07-30
Rename aiohabit project to mrack

Tiboris:
Rename aiohabit project to mrack


pvoborni:
Also LGTM. I've fixed Azure pipelines after the rename and  it aslo looks good.

PR #16, opened 2020-07-30
Load the requirements.txt in setup.py

Tiboris:
Load the content of requirements.txt file
instead of having static list in setup.py

Resolves https://github.com/pvoborni/mrack/issues/15 

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,0 +1,2 @@
+include LICENSE
+include requirements.txt
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -23,7 +23,7 @@ steps:
     python -m pip install --upgrade pip
     pip install -r requirements.txt
     pip install -r test-requirements.txt
-    pip install pytest pytest-azurepipelines flake8 black pydocstyle isort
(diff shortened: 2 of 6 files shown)


netoarmando:
Should we also include the packages from: 
https://github.com/pvoborni/mrack/blob/8ec7bf98391adb4bc02de381268577f857bbdd0a/azure-pipelines.yml#L26

Maybe as `test` or `dev` requirements.

pvoborni:
> Should we also include the packages from:
> 
> https://github.com/pvoborni/mrack/blob/8ec7bf98391adb4bc02de381268577f857bbdd0a/azure-pipelines.yml#L26
> 
> Maybe as `test` or `dev` requirements.

We should probably add them to test-requirements.txt But my "Python good practices knowledge" is limited here.

Tiboris:
> We should probably add them to test-requirements.txt But my "Python good practices knowledge" is limited here.

Yes I do agree. We should have all requirement to test locally in `test-requirements.txt` and if something more is needed in azure pipeline we should add it in azure's yaml.

netoarmando:
@Tiboris +1 to your proposal, I think you should include that to this PR.

PR #17, opened 2020-08-11
Handle output action error + Update logger format

netoarmando:
- Print message only
- Handle output action error
  - Running `mrack output` with an empty database breaks `python-pytest-multihost` config file. (E.g: Not running `mrack up` before trying that.)


pvoborni (approved):
Tested code looks good to me. Feel free to merge.

But I have to notes:
* the commit message mentions only one use case this fixed. The other: "Output command no longer fails when we have a DB but metadata contains host which the DB doesn't have. ". 
* when trying to reproduce the code path for ^^ I also found issue that it fails badly if metadata doesn't contain a host which the DB has. This can be handled in separate MR.
* also we fail "horribly" when metadata host is not dict.

PR #20, opened 2020-08-27
Add beaker provider

Tiboris:
Add beaker provider support

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,3 +1,4 @@
+beaker-client>=27
 boto3
 click
 pyyaml
--- a/src/mrack/providers/beaker.py
+++ b/src/mrack/providers/beaker.py
@@ -1,0 +1,331 @@
+# Copyright 2020 Red Hat Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Beaker Provider interface."""
+
+import asyncio
+import logging
+import os
+import socket
+import xml.etree.ElementTree as eTree
+
+from copy import deepcopy
+from datetime import datetime
+from bkr.common.hub import HubProxy
+from bkr.common.pyconfig import PyConfigParser
+from bkr.client import BeakerRecipeSet, BeakerRecipe, BeakerJob
+from xml.dom.minidom import Document as xml_doc
+
+
+from mrack.host import (
+    STATUS_ACTIVE,
+    STATUS_DELETED,
+    STATUS_ERROR,
+    STATUS_OTHER,
+    STATUS_PROVISIONING,
+    Host,
+)
+from mrack.providers.provider import Provider
+
+logger = logging.getLogger(__name__)
+
+PROVISIONER_KEY = "beaker"
+
+STATUS_MAP = {
+    "Reserved": STATUS_ACTIVE,
+    "New": STATUS_PROVISIONING,
+    "Processed": STATUS_PROVISIONING,
+    "Waiting": STATUS_PROVISIONING,
+    "Installing": STATUS_PROVISIONING,
+    "Running": STATUS_PROVISIONING,
+    "Cancelled": STATUS_DELETED,
+    "Aborted": STATUS_ERROR,
+    "Completed": STATUS_OTHER,
+}
+
+
+class BeakerProvider(Provider):
+    """Beaker Provider."""
+
+    def __init__(self):
+        """Object initialization."""
+        self._name = PROVISIONER_KEY
+        self.conf = PyConfigParser()
+        self.poll_sleep = 30  # seconds
+
+    async def init(self, prov_config):
+        """Initialize provider with data from Beaker configuration."""
+        self.prov_config = prov_config
+        logger.info("Initializing Beaker provider")
+        # eg: 240 attempts * 30s timeout - 2h timeout for job to complete
+        self.max_attempts = prov_config["max_attempts"]
+        login_start = datetime.now()
+        default_config = os.path.expanduser(
+            os.environ.get("BEAKER_CONF", "/etc/beaker/client.conf")  # TODO use provc
+        )  # get the beaker config for initialization of hub
+        self.conf.load_from_file(default_config)
+        self.hub = HubProxy(logger=logger, conf=self.conf)
+        login_end = datetime.now()
+        login_duration = login_end - login_start
+        logger.info(f"Login duration {login_duration}")
+
+    async def validate_hosts(self, hosts):
+        """Validate that host requirements are well specified."""
+        return
+
+    def _allow_ssh_key(self, ssh_key):
+
+        with open(ssh_key, "r") as key_file:
+            key_content = key_file.read()
+
+        return [
+            """%%post
+mkdir -p /root/.ssh
+cat >>/root/.ssh/authorized_keys << "__EOF__"
+%s__EOF__
+restorecon -R /root/.ssh
+chmod go-w /root /root/.ssh /root/.ssh/authorized_keys
+%%end"""
+            % "".join(key_content)
+        ]
+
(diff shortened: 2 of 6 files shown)


src/mrack/providers/beaker.py
```diff
@@ -0,0 +1,334 @@
...
+
+            if STATUS_MAP.get(status) == STATUS_ACTIVE:
+                break
+            elif STATUS_MAP.get(status) == STATUS_PROVISIONING:
+                await asyncio.sleep(self.poll_sleep)
+            elif STATUS_MAP.get(status) in [STATUS_ERROR, STATUS_DELETED]:
+                logger.warning(
+                    f"Job {beaker_id} has errored with status "
+                    f"{status} with result {resource['result']}"
+                )
```
pvoborni, line 249: Shouldn't here be also break? And probably in the case below as well. Otherwise it is iterating without a sleep.

src/mrack/providers/beaker.py
```diff
@@ -0,0 +1,334 @@
...
+        recipe.addBaseRequires(**specs)
+
+        # Specify the architecture
+        arch_node = xml_doc().createElement("distro_arch")
+        arch_node.setAttribute("op", "=")
+        arch_node.setAttribute("value", specs["arch"])
+        recipe.addDistroRequires(arch_node)
+
+        # Add ReserveSys element to reserve system after provisioning
+        recipe.addReservesys(duration="86400")
```
pvoborni, line 138: In the first version seems OK to have it hardcoded but later we will probably need to have reservation duration configurable.

src/mrack/providers/beaker.py
```diff
@@ -0,0 +1,334 @@
...
+        Issues provisioning and waits for it succeed. Raises exception if any of
+        the servers was not successfully provisioned. If that happens it issues deletion
+        of all already provisioned resources.
+
+        Return list of information about provisioned servers.
+        """
+        started = datetime.now()
+
+        count = len(hosts)
+        logger.info(f"Issuing provisioning of {count} hosts")
```
pvoborni, line 271: A Q for future thoughs, should we log messages like this with some provider prefix? So that we can see in log e.g.:

```
OpenStack: Issuing provisioning of 3 hosts
Beaker: Issuing provisioning of 2 hosts
```

Instead of:
```
Issuing provisioning of 3 hosts
Issuing provisioning of 2 hosts
```

pvoborni (commented):
Hi Tibor, thanks for this PR. I did first only read, quick look, high level review with some comments.

src/mrack/providers/beaker.py
Tiboris: Yes that would be nice!

src/mrack/providers/beaker.py
Tiboris: My bad :)

src/mrack/providers/beaker.py
Tiboris: Yes that's correct

Tiboris:
@netoarmando could you please have a look at https://dev.azure.com/pvoborni/mrack/_build/results?buildId=78&view=logs&j=00b6a206-eb91-585b-3d14-cf1a4d7b1970&t=b5464ffe-109e-5907-849b-650dd16a8b4d&l=16

netoarmando:
> @netoarmando could you please have a look at https://dev.azure.com/pvoborni/mrack/_build/results?buildId=78&view=logs&j=00b6a206-eb91-585b-3d14-cf1a4d7b1970&t=b5464ffe-109e-5907-849b-650dd16a8b4d&l=16

Here's the fix: https://github.com/pvoborni/mrack/pull/21

requirements.txt
```diff
@@ -1,3 +1,4 @@
+beaker-client==27.4
```
netoarmando, line 1: It would be nice to improve the commit message to include some reasoning about your decisions. (e.g. why setting the the version)

src/mrack/providers/beaker.py
netoarmando: We can use filters for that (https://docs.python.org/3/howto/logging-cookbook.html#filters-contextual) in case we don't want to print that explicitly.

Tiboris: Let us go this way

requirements.txt
Tiboris: I tested this with versions 27.4 and 28 in requirements i will leave >=27

src/mrack/transformers/beaker.py
```diff
@@ -0,0 +1,62 @@
...
+from mrack.utils import get_config_value, print_obj
+
+CONFIG_KEY = "beaker"
+
+
+class BeakerTransformer(Transformer):
+    """Beaker transformer."""
+
+    _config_key = CONFIG_KEY
+    _required_config_attrs = ["distros", "client_conf", "keypair"]
```
netoarmando, line 28: Attribute `client_conf` is not mentioned anywhere, what is that?

src/mrack/providers/beaker.py
```diff
@@ -0,0 +1,333 @@
...
+
+        provisioned = datetime.now()
+        provi_duration = provisioned - started
+
+        logger.info("All Beaker hosts reached provisioning final state (Reserved)")
+        logger.info(f"Provisioning duration: {provi_duration}")
+
+        hosts = [self.to_host(srv) for srv in server_results]
+        for host in hosts:
+            logger.info(host)
```
netoarmando, line 295: Output of my test was:
```
Provisioning duration: 0:00:38.012023
error J:4536747 f30.aiohabit.test 10.XXX.XXX.6 None None
Error:
"Warn"
```
Last two lines came from Beaker client but they don't say a lot, did you see the same thing?

src/mrack/providers/beaker.py
```diff
@@ -0,0 +1,333 @@
...
+            host.id, "cancel", "Job has been cancelled by mrack."
+        )
+
+    async def delete_hosts(self, hosts):
+        """Issue deletion of all servers based on previous results from provisioning."""
+        logger.info("Issuing Beaker deletion")
+
+        delete_bkr = []
+        for host in hosts:
+            aws = self.delete_host(host)
```
netoarmando, line 312: nitpick: I would avoid `aws` just because we also use those letters to refer to "AWS®", making the output of `grep` a little noisy.

pvoborni: agree that it is noisy and also bit confusing, what var name would you suggest? It is a shortcut for awaitables. Quite used in https://docs.python.org/3/library/asyncio-task.html

src/mrack/transformers/beaker.py
Tiboris: It should be bkr client conf

src/mrack/providers/beaker.py
Tiboris: Let us investigate this

src/mrack/providers/beaker.py
Tiboris: Fixed

src/mrack/transformers/beaker.py
Tiboris: I have removed that

src/mrack/providers/beaker.py
Tiboris: This was caused because the distro form metadata was not in prov-config and beaker selected default distro.
Beaker failed with: 
`Failed to provision recipeid 8753124, Failed to find repo for harness `
I will file an issue with that to have better handling for such cases.

netoarmando: As discussed via IRC: This happened because I've requested a "wrong" distribution.

With a proper metadata.yaml file everything went well.
The error is still there for "wrong" scenarios. Perhaps this must be addressed by the `validate_hosts()` method, but we it can be done in a future PR.

src/mrack/transformers/beaker.py
```diff
@@ -0,0 +1,60 @@
...
+
+class BeakerTransformer(Transformer):
+    """Beaker transformer."""
+
+    _config_key = CONFIG_KEY
+    _required_config_attrs = ["distros", "keypair", "reserve_duration", "max_attempts"]
+
+    async def init_provider(self):
+        """Initialize associate provider."""
+        await self._provider.init(self.config)
```
netoarmando, line 32: Note for the future: I wonder how tightly we are coupling XProvider with XTransformer. The `init` method for `AWSProvider` and `OpenStackProvider` is `async def init(self, image_names=None)`.

It's not a blocker here, based on my understanding on how mrack is currently working, but perhaps something we need to consider eventually.

netoarmando (approved):
Everything looks good to me, there were only some notes we should consider for future steps.

PR #21, opened 2020-08-27
Fix style after black update

netoarmando:
Latest version of black was complaing about these changes.

Previous version: black-19.10b0
Latest version: black-20.8b1


Tiboris:
Thanks! Lets merge this.

PR #24, opened 2020-09-09
Use public AWS addresses instead of private ones and add some minor fixes to aws

Tiboris:
Using private addresses will result into unreachable host
without vpn tunnel or mechanism to access AWS network.

Add security group, tags, keypair to init of AWS provider.
They have to be defined in provisioning config.

Remove asserts and replace them with ProvisioningErrors

### DIFF
--- a/src/mrack/providers/aws.py
+++ b/src/mrack/providers/aws.py
@@ -21,6 +21,7 @@ from datetime import datetime
 
 import boto3
 
+from mrack.errors import ProvisioningError
 from mrack.host import (
     STATUS_ACTIVE,
     STATUS_DELETED,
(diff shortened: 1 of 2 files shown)


src/mrack/providers/aws.py
```diff
@@ -101,20 +98,20 @@ async def create_server(self, req):
...
             InstanceType=specs.get("flavor"),
+            KeyName=self.ssh_key,
+            SecurityGroupIds=[self.sec_group],
         )
 
         ids = [srv.id for srv in aws_res]
-        assert len(ids) == 1  # ids must be len of 1 as we provision one vm at the time
-
+        if len(ids) != 1:  # ids must be len of 1 as we provision one vm at the time
+            raise (ProvisioningError)
```
netoarmando, line 107:
```python
raise ProvisioningError("Unexpected number of instances provisioned.")
```
_That is just a suggestion, feel free to improve the message._

src/mrack/providers/aws.py
```diff
@@ -146,11 +140,11 @@ async def wait_till_provisioned(self, aws_id):
...
-        assert len(response["Reservations"]) == 1
-        assert len(response["Reservations"][0]["Instances"]) == 1
-        result = response["Reservations"][0]["Instances"][0]
-        # returns dict with aws instance information
-        return result
+
+        try:  # returns dict with aws instance information
+            return response["Reservations"][0]["Instances"][0]
+        except (KeyError, IndexError):
+            raise (ProvisioningError)
```
netoarmando, line 147:
```python
raise ProvisioningError("Missing data provisioned instance.")
```
_That is just a suggestion, feel free to improve the message._

netoarmando (commented):
Nice improvements! Thanks for that. Please, consider the suggestions I left when raising exceptions.

src/mrack/providers/aws.py
Tiboris: Will do :)

src/mrack/providers/aws.py
Tiboris: Sure!

netoarmando (approved):

PR #26, opened 2020-09-10
Add username='root' to fix ansible user

Tiboris:
Add username='root' to fix ansible user

Added username='root' to beaker host object initialization
so inventory is created using ansible_user: root.
This is required because we are uploading public key
while provisioning to the root user and not user set
in the provisioning config.
The mismatched user causes unreachable host
when using generated ansible inventory.

Fix initialization of provider in a way
that we do not pass provisioning config
but only needed values.

Add status Queued as provisioning.

### DIFF
--- a/src/mrack/providers/beaker.py
+++ b/src/mrack/providers/beaker.py
@@ -19,14 +19,13 @@ import logging
 import os
 import socket
 import xml.etree.ElementTree as eTree
-
 from copy import deepcopy
 from datetime import datetime
-from bkr.common.hub import HubProxy
-from bkr.common.pyconfig import PyConfigParser
-from bkr.client import BeakerRecipeSet, BeakerRecipe, BeakerJob
 from xml.dom.minidom import Document as xml_doc
 
+from bkr.client import BeakerJob, BeakerRecipe, BeakerRecipeSet
+from bkr.common.hub import HubProxy
+from bkr.common.pyconfig import PyConfigParser
 
 from mrack.host import (
     STATUS_ACTIVE,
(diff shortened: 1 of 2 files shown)


src/mrack/providers/beaker.py
```diff
@@ -112,9 +113,8 @@ def _req_to_bkr_job(self, req):
         specs.update({"priority": "Normal"})
 
         # Add allowed keys
-        keypair = self.prov_config.get("keypair", None)
-        if keypair:
-            specs.update({"ks_append": self._allow_ssh_key(keypair)})
+        if self.keypair:
```
netoarmando, line 116: `keypair` is listed in the `_required_config_attrs` property for Beaker, is this conditional still necessary for some reason?

src/mrack/providers/beaker.py
```diff
@@ -63,13 +62,15 @@ def __init__(self):
         self._name = PROVISIONER_KEY
         self.conf = PyConfigParser()
         self.poll_sleep = 30  # seconds
+        self.keypair = None
 
-    async def init(self, prov_config):
+    async def init(self, max_attempts, reserve_duration, keypair):
```
netoarmando, line 67: :+1:

netoarmando (commented):
Please also add to the commit message a super quick reason why `ansible_user: root` is necessary (e.g.: "Beaker images requirement")

src/mrack/providers/beaker.py
Tiboris: eh no you are correct.

src/mrack/providers/beaker.py
Tiboris: I am adding another status because queued seems to be missing and causes failure when beaker is busy.

netoarmando (approved):

PR #29, opened 2020-09-13
ansible-inventory: rework generation of groups

pvoborni:
The Ansible inventory output was not correctly adding hosts into groups. E.g.
group was not added when it was not defined in layout. Nested group was not
correctly found and host added.

This part is completely reworked and changes the behavior to:
- add only a host name reference (no vars) into custom groups
- add missing groups as children of 'all' group
- works with both "groups" and "group" defined in metadata
- add all hosts to 'all' with also variables. This ensures that the full host
  object is defined only once.

Doc: https://docs.ansible.com/ansible/latest/user_guide/intro_inventory.html

### DIFF
--- a/src/mrack/outputs/ansible_inventory.py
+++ b/src/mrack/outputs/ansible_inventory.py
@@ -16,6 +16,7 @@
 
 from copy import deepcopy
 
+from mrack.errors import ConfigError
 from mrack.outputs.utils import resolve_hostname
 from mrack.utils import (
     get_host_from_metadata,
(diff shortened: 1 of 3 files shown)


tests/unit/test_ansible_inventory.py
```diff
@@ -0,0 +1,77 @@
...
+        self.db = get_db_from_metadata(self.metadata)
+
+    def test_common_layout(self):
+
+        layout = common_inventory_layout()
+        config = provisioning_config(layout)
+
+        ans_inv = AnsibleInventoryOutput(config, self.db, self.metadata)
+
+        inventory = ans_inv.create_inventory()
```
netoarmando, line 69: You can move this to a fixture called `inventory`; (You can do the same to `metadata` and `db`) Pass them to `test_common_layout` and move the `ensure_*` functions to become methods.

Fixtures are the best practice to setup and teardown scenarios.

pvoborni: Amended and several tests added (through parametrize).

src/mrack/outputs/ansible_inventory.py
```diff
@@ -38,30 +39,61 @@ def copy_meta_attrs(host, meta_host, attrs):
...
-        if "children" not in g:
-            continue
+def add_to_group(inventory, groupname, hostname):
+    """
+    Find a group in inventory layout and add a host to it.
 
-        found = find_group(g["children"], group_names, top=False)
-        if found:
-            return found
+    Returns True if host was added to a group, False if not.
```
Tiboris, line 77: Just a note that it does not return `False`/`True` Could you please update the doctext?

Tiboris (changes requested):
Thanks for the update @pvoborni changes looks good to me, but please check the inline comment.
Let us wait for @netoarmando to give final review.

netoarmando (approved):
LGTM! I believe the solution for parametrizing the fixtures is good enough!

Tiboris (approved):

PR #31, opened 2020-09-18
Refactor providers to share same code

Tiboris:
Methods provision_hosts, delete_hosts and to_host
were moved to provider.py as they can be easily
shared among the providers.

Added missing methods to provider.py to share
same structure. Not implemented methods will throw
a NotImplementedError() and they are ment to be
implmented per provider.

Moved host count to initialization of openstack
provider to set poll sleeps based on this.

For all providers moved STATUS_MAP to class
so from now on it is a class attribute.

Openstack's wait_till_provisioned now uses
instance and not only id to match inherited method.

Added optional username parameter to to_host method.

### DIFF
--- a/src/mrack/providers/openstack.py
+++ b/src/mrack/providers/openstack.py
@@ -23,14 +23,13 @@ from urllib.parse import parse_qs, urlparse
 from asyncopenstackclient import AuthPassword, GlanceClient
 from simple_rest_client.exceptions import NotFoundError
 
-from mrack.errors import ProvisioningError, ServerNotFoundError, ValidationError
+from mrack.errors import ServerNotFoundError, ValidationError
 from mrack.host import (
     STATUS_ACTIVE,
     STATUS_DELETED,
     STATUS_ERROR,
     STATUS_OTHER,
     STATUS_PROVISIONING,
-    Host,
 )
 from mrack.providers.provider import Provider
 from mrack.providers.utils.osapi import ExtraNovaClient, NeutronClient
@@ -43,16 +42,6 @@ logger = logging.getLogger(__name__)
 
 
 PROVISIONER_KEY = "openstack"
-
-
-STATUS_MAP = {
-    "ACTIVE": STATUS_ACTIVE,
-    "BUILD": STATUS_PROVISIONING,
-    "DELETED": STATUS_DELETED,
-    "ERROR": STATUS_ERROR,
-    # there is much more we can treat it as STATUS_OTHER, see:
-    # https://docs.openstack.org/api-guide/compute/server_concepts.html
-}
 
 
 class OpenStackProvider(Provider):
@@ -78,6 +67,14 @@ class OpenStackProvider(Provider):
         self.timeout = 60  # minutes
         self.poll_sleep_initial = 15  # seconds
         self.poll_sleep = 7  # seconds
+        self.STATUS_MAP = {
+            "ACTIVE": STATUS_ACTIVE,
+            "BUILD": STATUS_PROVISIONING,
+            "DELETED": STATUS_DELETED,
+            "ERROR": STATUS_ERROR,
+            # there is much more we can treat it as STATUS_OTHER, see:
+            # https://docs.openstack.org/api-guide/compute/server_concepts.html
+        }
 
     async def init(self, image_names=None):
         """Initialize provider with data from OpenStack.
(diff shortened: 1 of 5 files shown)


src/mrack/providers/provider.py
```diff
@@ -30,16 +37,116 @@ def name(self):
...
+            wait_servers.append(awaitable)
+
+        server_results = await asyncio.gather(*wait_servers)
+        provisioned = datetime.now()
+        provi_duration = provisioned - started
+
+        logger.info("All hosts reached provisioning final state (ACTIVE or ERROR)")
+        logger.info(f"Provisioning duration: {provi_duration}")
+
+        errors = [res for res in server_results if res["status"] == "ERROR"]
```
pvoborni, line 105: This status field and it's values are OpenStack specific. Part of `server_result` object can be unified between providers.

src/mrack/providers/provider.py
```diff
@@ -30,16 +37,116 @@ def name(self):
...
+        logger.info("Issuing deletion")
+        delete_servers = []
+        for host in hosts:
+            awaitable = self.delete_host(host)
+            delete_servers.append(awaitable)
+        results = await asyncio.gather(*delete_servers)
+        logger.info("All servers issued to be deleted")
+        return results
+
+    def _get_host_info_from_prov_result(self, prov_result):
```
pvoborni, line 134: Seems to me that length of this method is longer than `to_host` in all/almost all providers. It seems cleaner to me to have in base class:
```
def to_host(self, provisioning_result):
        """Transform provisioning result into Host object."""
       raise NotImplementedError()
```

and have content of `_get_host_info_from_prov_result` in `to_host`:

pvoborni (commented):
Like it, but needs bit more work.

src/mrack/providers/provider.py
netoarmando: I agree, also it's odd to require a 'private' method to be implemented.

With this change, please, consider to group all methods that require implementation before the one implemented by the base class.

src/mrack/providers/provider.py
```diff
@@ -30,16 +37,116 @@ def name(self):
...
+        provider has enough resources(quota).
+
+        Then issues provisioning and waits for it succeed. Raises exception if any of
+        the servers was not successfully provisioned. If that happens it issues deletion
+        of all already provisioned resources.
+
+        Return list of information about provisioned servers.
+        """
+        if not logger:
+            logger = logging.getLogger(self.name)
```
netoarmando, line 69: I guess you want to pass the `logger` instance to have a 'proper' prefix, right? I _think_ this goes against some of the assumptions the logging makes.

pvoborni:
The following commit contains the idea which I described in the review comments. Only for OpenStack provider. Untested.

https://github.com/pvoborni/mrack/commit/71bef84bf9e543bdca8f35ef5de717408abf1c96

src/mrack/providers/provider.py
Tiboris: Got it i will try to update this asap

src/mrack/providers/provider.py
netoarmando: Final solution is okay to me, one public method responsible for translation the data and we keep `def to_host` as the `Host` instances factory for now.

src/mrack/providers/provider.py
Tiboris: I have added parse_errors method.

src/mrack/providers/openstack.py
```diff
@@ -90,7 +87,10 @@ async def init(self, image_names=None):
         * account limits (max and current usage of vCPUs, memory, ...)
         """
         # session expects that credentials will be set via env variables
+        if host_cnt == 0:
+            raise ProvisioningError("Provider initialized with  0 required host count")
         logger.info("Initializing OpenStack provider")
+        self.poll_sleep_initial, self.poll_sleep = self.get_poll_sleep_times(host_cnt)
```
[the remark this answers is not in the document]
Tiboris: I have removed the get_poll_sleep_times let us resolve it this become a issue.

src/mrack/providers/provider.py
netoarmando: Final solution is "no-matter-the-provider-proof", in the future we can solve the way we handle the status everywhere.

netoarmando (approved):

PR #36, opened 2020-10-05
Add mrack.conf which contains default enfironment

Tiboris:
Add mrack.conf from which we load defaults

Add mrack.conf which contains default enfironment
configuration in ini format for the click options
- `--config=provisioning-config from mrack.conf`
- `--db=mrackdb from mrack.conf`

Load default for options from the mrack.conf file from:
1. use `$(pwd)/mrack.conf` or other way the `./mrack.conf` - actual directory
2. or use user home directory configuration - `~/.mrack/mrack.conf`
3. alternatively directory `/etc/mrack/mrack.conf` as safe fallback.

Add example `mrack.conf` and `provisioning-config.yaml`
to the `data/` directory in repository.

Update setup.py to install required files from `data/` dir.

### DIFF
--- a/src/mrack/data/provisioning-config.yaml
+++ b/src/mrack/data/provisioning-config.yaml
@@ -1,0 +1,82 @@
+aws: # aws provider config
+    images:
+        # list of ami images key is used in metadata ami image is used for provider e.g.
+        fedora-32: ami-0f1d60751946b66c5 # Fedora-Cloud-Base-32-1.6.x86_64-hvm-eu-central-1-gp2-0
+
+
+    flavors:
+        # list of available flavours to ask from provider  for vm specs
+        default: t2.nano # e.g. this is 1vcpus 0.5GB ram
+
+    keypair: mrack # this name points to uploaded and available in aws ec2 key configuration
+    security_group: sg-0ad1bab3a850e70fc # setup the system firewall using one of existing sec groups
+
+    credentials_file: aws.key # file containing the credentials
+    profile: default # credentials profile to use
+    region: eu-central-1 # default region for ec2 instances
+    instance_tags: # custom tag list to add to vm created using mrack
+        Name: "mrack-runner"
+        mrack: "True"
+        persistent: "False"
+
+
+beaker: # beaker provider specific values
+    distros:
+        # list of beaker distros key is used in metadata distro is used in generated jobxml
+        fedora-32: Fedora-32%
+
+    # path to public ssh key which is later uploaded to the machine for access
+    keypair: id_rsa.pub
+    # default reservation time for the beaker job
+    reserve_duration: 86400
+    # default max_attepmts value for the beaker job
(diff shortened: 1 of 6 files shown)


pvoborni:
The code seems to expect that the config will be placed in a system directory. This can be a possibility but it doesn't fulfill the original use case. The config file should be loaded from the directory where mrack is executed - in IdM CI terminology - the TWD. If it is not present then it will be ignored. Alternatively mrack can try other locations like `~/.config/mrack` and the system one as a last option. In either case it should work without the file present.

src/mrack/data/provisioning-config.yaml
```diff
@@ -0,0 +1,117 @@
+aws:
```
pvoborni, line 1: I don't think we should have a default provisioning config with so many values specific to certain environment. Basically it should not contain anything which is not universal.

src/mrack/run.py
```diff
@@ -43,6 +46,58 @@
...
+def get_config_file_path():
+    """Locate mrack.conf file."""
+    try:
+        return os.path.join(get_mrack_dir(), "mrack.conf")
+    except BrokenInstallation:
+        logger.warning("Can not load mrack.conf using Python package defaults.")
+        return os.path.join(os.path.dirname(__file__), "data/mrack.conf")
+
+
+def get_default_file_path(filename):
```
pvoborni, line 73: Why not to work with key names directly? What is the benefit of using key names similar to some default values when you will get an error anyway if you specify a different filename when calling the method.

I.e. the method seems to me too complex without  a clear benefit.

src/mrack/data/provisioning-config.yaml
netoarmando: I'm inclined to agree, does that mean we need "mostly" 1 image, 1 distro, 1 flavor etc?
Some projects have a default config file with lots of comment blocks, explaining every parameter with one (or none) default value enabled. As a user I would appreciate that.

pvoborni: +1, commented examples could be good.

Tiboris: @netoarmando Will do so

pvoborni (commented):
Originally I wanted to approve this with minor comments. But the more I tried it and e.g. found, that writing metadata is still also inconvenient. I started to think more about other use cases and the implementation. And in the end tried a different approach  in https://github.com/neoave/mrack/pull/55 due to the following reasons:
- the get_file_path_from_config method is too overloaded - it does 3 separate things - constructing list of possible paths,  loading the config, finding a value
- is inefficient - it is doing the file operations again and again for each key to load
- hard to extend - it is focusing only on paths, not thinking in a future that we want to store other configuration. E.g. debug level, provisioning strategy or default provisioner.
- due to it doing a lot of stuff it is also hard to read and understand quickly

PR #41, opened 2020-10-14
docs: set master doc to index

pvoborni:
From sphinx doc:
```
master_doc

    The document name of the “master” document, that is, the document
    that contains the root toctree directive. Default is 'index'.

    Changed in version 2.0: The default is changed to 'index' from 'contents'.
```

Read the docs is using version < 2.0 thus it tries to find 'content's
and it fails.

Also changing theme to default - no real reason, thinking it will be
better.

Currently build fails with: 
`master file /home/docs/checkouts/readthedocs.org/user_builds/mrack/checkouts/latest/docs/contents.rst not found`

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -24,6 +24,8 @@ author = "Petr Vobornik"
 # The full version, including alpha/beta/rc tags
 release = "0.2.0"
 
+# to work with ReadTheDocs which is using version < 2.0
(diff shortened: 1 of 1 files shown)


pvoborni:
@netoarmando this helped. See https://mrack.readthedocs.io/en/latest/# So now we should add content :) . CC @Tiboris

PR #42, opened 2020-10-26
Remove missing and unused print_basic_info method

Tiboris:
Remove missing and unused print_basic_info method
    
    Add loger info message to error parsing method.
    Move the parse_errors to provider.py file.
    Add property error to Host object.
    Delete host now needs only host_id.

### DIFF
--- a/src/mrack/providers/provider.py
+++ b/src/mrack/providers/provider.py
@@ -18,7 +18,7 @@ import logging
 from datetime import datetime
 
 from mrack.errors import ProvisioningError, ValidationError
-from mrack.host import STATUS_OTHER, Host
+from mrack.host import STATUS_ERROR, STATUS_OTHER, Host
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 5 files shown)


src/mrack/providers/provider.py
```diff
@@ -101,25 +101,39 @@ async def provision_hosts(self, hosts):
...
-        raise NotImplementedError()
+        errors = []
+        for host in hosts:
+            logger.info(f"Host: {host.id}\tStatus: {host.status}")
+            if self.STATUS_MAP.get(host.status, STATUS_OTHER) == STATUS_ERROR:
+                errors.append(host.error)
+
+        return errors
+
+    def print_err_info(self, err):
```
netoarmando, line 132: In the end, do we really need this method?

pvoborni: In the beginning. parse_errors method was not about extracticting error messages. But it was about getting the provisioning results where host were not provisioned correctly so that we can decide what to do with it. The implemented default method was to print the error information (thus the print_err_info method) and delete all provisioned hosts. But a different provisioning strategy can delete just the hosts in error and re-try their provisioning.

Tiboris: I have removed this method for now.

netoarmando (approved):

PR #43, opened 2020-10-28
Add docs on how to set credentials

netoarmando:
Quick guide on how set credentials for current supported providers.


pvoborni:
Did a quick read and it looks good to me. Though I don't say that I followed the instructions so if it is bad then I did not noticed that :).

pvoborni:
I wonder if we can somehow build the documentation in PRs, so that we can also check if it looks good.

netoarmando:
> I wonder if we can somehow build the documentation in PRs, so that we can also check if it looks good.

We can, but when I ran `make html` with a typo instead the correct path to `credentials.rst`, the output was helpful to detect the error; however, the return code was always `0`.

pvoborni:
One thing is pass the build. This should be added to Azure. But I also though, if e.g. read the docs or something else can build the branch so that we can see how it actually looks, e.g. if it is correctly formatted - without a need to check it out locally, build it and see it.

pvoborni:
In any case, I don't have any problem with merging this PR.

docs/guides/credentials.rst
```diff
@@ -0,0 +1,66 @@
+How to configure provider's credentials
+=======================================
+
+mrack supports several providers by using popular libraries without overriding
+how credentials are handled. That means every provider must be configured
+differently and according to its library.
+
+This guide explains how you can configure those credentials including pointers
+to original resources you can find more info about that.
```
Tiboris, line 9:
```suggestion
to original resources where you can find more info about that.
```

docs/guides/credentials.rst
```diff
@@ -0,0 +1,66 @@
...
+
+AWS
+---
+mrack uses ``boto3`` library to interact with AWS, thus you can follow the
+instructions from Boto's `documentation <https://boto3.amazonaws.com/v1/documentation/api/latest/guide/credentials.html#guide-credentials>`__
+
+The most straightforward way is by using a shared config file. Its default location
```
Tiboris, line 16:
```suggestion
The most straightforward way is to use a shared config file. Its default location
```

docs/guides/credentials.rst
```diff
@@ -0,0 +1,66 @@
...
+---------
+Before calling ``mrack`` you must define some environment variables or source the
+OpenStack RC file provided by your instance.
+
+.. code:: bash
+
+   $ source myproject-openrc.sh
+   $ mrack up
+
+Instructions on where you can file that RC file are available `here <https://docs.openstack.org/mitaka/cli-reference/common/cli_set_environment_variables_using_openstack_rc.html>`__
```
Tiboris, line 39: Did you mean find here?
```suggestion
Instructions on where you can find the RC file are available `here <https://docs.openstack.org/mitaka/cli-reference/common/cli_set_environment_variables_using_openstack_rc.html>`__
```

docs/guides/credentials.rst
```diff
@@ -0,0 +1,66 @@
...
+           fedora-32: Fedora-32%
+
+       client_conf: /etc/beaker/client.conf
+       keypair: id_rsa.pub
+
+       reserve_duration: 86400
+       server_role: idm_ci_bkr_server
+       max_attempts: 240
+
+More info about Beaker client configuration available `here <https://beaker-project.org/docs/user-guide/bkr-client.html>`__
```
Tiboris, line 61:
```suggestion
More info about Beaker client configuration is available `here <https://beaker-project.org/docs/user-guide/bkr-client.html>`__
```

docs/index.rst
```diff
@@ -1,15 +1,66 @@
-.. iohabit documentation master file, created by
+.. mrack documentation master file, created by
```
Tiboris, line 1: ouch :smile:

docs/index.rst
```diff
@@ -1,15 +1,66 @@
...
+       IPv6: net_ipv6_only
+       dual: net_cci_4_6
+     keypair: ipa_key
+   beaker:
+     distros:
+       fedora-30: FEDORA-30%
+       fodora-31: FEDORA-31%
+
+So then user can quickly change provisioning location just by changing
+provider name without needing to redefine the provider/cloud specifics.
```
Tiboris, line 54:
```suggestion
provider name without a need to redefine the provider/cloud specifics.
```

Tiboris (changes requested):
@netoarmando Thanks for the patch!

There are some nitpicks which may be good to at least look at I do not insist on change but some changes seems to be in place.
My english no perfect but I have 57 to 78 certicates :laughing:  :wink: .

Please have a look on the comments.

netoarmando:
@pvoborni @Tiboris I added a basic support to build docs and save them into Azure's artifacts.

Tiboris (approved):
Thanks @netoarmando !

Issue #23, opened 2020-09-02
Give context to output


netoarmando:
Goal is to print this:
```
OpenStack: Issuing provisioning of 3 hosts
Beaker: Issuing provisioning of 2 hosts
```

Instead of:
```
Issuing provisioning of 3 hosts
Issuing provisioning of 2 hosts
```

Logging filters can be used for that (https://docs.python.org/3/howto/logging-cookbook.html#filters-contextual)

---

_This was discussed at: https://github.com/pvoborni/mrack/pull/20#discussion_r478353983_

PR #45, opened 2020-10-30
Improve mrack logging

Tiboris:
refactor: Add dsp_name to logging messages to display context
    
    Add dsp_name to logging messages to display context
    for prioviders and transformers.
    Added some debug messages.
    
Aims to resolve https://github.com/neoave/mrack/issues/23 

### DIFF
--- a/src/mrack/providers/aws.py
+++ b/src/mrack/providers/aws.py
@@ -36,7 +36,7 @@ class AWSProvider(Provider):
     def __init__(self):
         """Object initialization."""
         self._name = PROVISIONER_KEY
-        self.display_name = "AWS"
+        self.dsp_name = "AWS"
         self.ami_ids = []
         self.ssh_key = None
         self.sec_group = None
@@ -58,13 +58,13 @@ class AWSProvider(Provider):
     async def init(self, ami_ids, ssh_key, sec_group, instance_tags):
         """Initialize provider with data from AWS."""
         # AWS_CONFIG_FILE=`readlink -f ./aws.key`
-        logger.info("Initializing AWS provider")
+        logger.info(f"{self.dsp_name}: Initializing provider")
         login_start = datetime.now()
         self.ec2 = boto3.resource("ec2")
         self.client = boto3.client("ec2")
         login_end = datetime.now()
         login_duration = login_end - login_start
-        logger.info(f"Login duration {login_duration}")
+        logger.info(f"{self.dsp_name}: Login duration {login_duration}")
         self.ami_ids = ami_ids
         self.ssh_key = ssh_key
         self.sec_group = sec_group
@@ -76,17 +76,22 @@ class AWSProvider(Provider):
             req_img = req.get("image")
             if not req.get("meta_image") and req_img not in self.ami_ids:
                 raise ValidationError(
-                    f"{self.display_name} provider does not support "
-                    f"'{req_img}' image in provisioning config."
+                    f"{self.dsp_name}: Provider does not support "
+                    f"'{req_img}' image in provisioning config"
                 )
 
             try:
                 aws_image = self.ec2.Image(req_img)
-                logger.info(f"Requested provisioning of {aws_image.name} image.")
+                logger.info(
+                    f"{self.dsp_name}: Requested provisioning of {aws_image.name} image"
+                )
             except ClientError as image_err:
-                err_resp = image_err.response["Error"]["Message"]
-                err_msg = f"Requested image '{req_img}' can not be provisioned."
+                err_msg = (
+                    f"{self.dsp_name}: Requested image "
+                    f"'{req_img}' can not be provisioned"
+                )
                 logger.error(err_msg)
+                err_resp = image_err.response["Error"]["Message"]
                 raise ValidationError(f"{err_msg} Request failed with {err_resp}")
 
         return
@@ -104,7 +109,7 @@ class AWSProvider(Provider):
         * 'image': ami or name of image
         * 'flavor': flavor to use
         """
-        logger.info("Creating AWS server")
+        logger.info(f"{self.dsp_name}: Creating server")
         specs = deepcopy(req)  # work with own copy, do not modify the input
 
         aws_res = self.ec2.create_instances(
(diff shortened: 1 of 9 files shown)


netoarmando (changes requested):
This is breaking logging for some reason, I tried to execute this:
```bash
mrack --debug -c non-existent_prov_config.yaml up  non-existent_metadata.yaml
```
No output was returned.

Tiboris:
> ```shell
> mrack --debug
> ```

yes you are correct when looking at this it should raise exception
https://github.com/neoave/mrack/blob/master/src/mrack/run.py#L194
and it raises it with the known_error when file is not found seems to be the:
<class 'mrack.errors.ConfigError'>
where it breaks, seems like exception_handler can not use logger with my changes.

Tiboris:
@netoarmando I hope it is fixed, reason for that can be seen in file `__init__.py`
https://github.com/neoave/mrack/pull/45/files#diff-3b06fe18a2bbf59c91431116c264ffac811c2e81bf485eed06ca51536c7e896cR15
The decorator does not pass the right function name for the logger.

src/mrack/logging.py
```diff
@@ -0,0 +1,63 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Logging interface."""
+import logging
+
+DEBUG = logging.DEBUG
```
netoarmando, line 18: I guess you are defining this so other parts can import that from mrack.logging, right?

How come the linter is not complaining about this?

src/mrack/transformers/transformer.py
```diff
@@ -72,3 +73,13 @@ def validate_host(self, host):
         provider = host.get("provider")
         if provider and provider not in providers.names:
             raise MetadataError(f"Error: Invalid host provider: {provider}")
+
+    def create_host_requirement(self, host):
+        """Create single input for provisioner."""
+        raise NotImplementedError()
+
+    def create_host_requirements(self):
```
netoarmando, line 81: nitpick: I would like to see these changes in another commit.

src/mrack/run.py
```diff
@@ -198,7 +198,7 @@ def handle(*args, **kwargs):
             ProviderError,
             ApplicationError,
         ) as known_error:
-            logger.error(known_error)
+            logger.error(known_error, extra={"name_override": func.__name__})
```
netoarmando, line 201: In the end this was necessary. Can't we leave the logger configuration on `src/mrack/__init__.py`, so if any module don't need to set a context can just use the logger with no surprises? The reasoning is that we could call logger from anywhere, without worrying about it breaking in production.

src/mrack/transformers/transformer.py
Tiboris: Will do.

src/mrack/run.py
Tiboris: Hmmm this should not be needed.

src/mrack/logging.py
Tiboris: Really, dunno :D maybe because it is upper-case.

pvoborni:
Imagine that you are a person who don't know the implementation of mrack. What is your feeling as the user when you are using mrack with this logging?

src/mrack/run.py
Tiboris: I would use only mrack logging so we do not mix up the loggers and the log is consistent.

src/mrack/transformers/transformer.py
Tiboris: Removed i will create another PR

Tiboris:
> Imagine that you are a person who don't know the implementation of mrack. What is your feeling as the user when you are using mrack with this logging?

Could you please collaborate more on what do you have in mind? I am afraid that I do not understand what you are trying to point out. :(

pvoborni:
> > Imagine that you are a person who don't know the implementation of mrack. What is your feeling as the user when you are using mrack with this logging?
> 
> Could you please collaborate more on what do you have in mind? I am afraid that I do not understand what you are trying to point out. :(

It's a question to you and Armando :) Don't want to point out anything yet.  Sort of a nudge to apply a perspective different from ours (developers).

netoarmando:
> Imagine that you are a person who don't know the implementation of mrack. What is your feeling as the user when you are using mrack with this logging?

For me, the ideal scenario would be to use python standard logging call, BUT with the option to use our internal logging for context.. However, I do think that we can refer to the Zen of Python, which **might** mean to return to plain loggers with explicit strings (e.g.: `f"{provider}: message"` because it is simpler and explicit.

pvoborni:
I actually never red Zen of Python, thanks for the reference. https://www.python.org/dev/peps/pep-0020/

In the spirit of it I'd add: "Experience for end users is more important than purity of implementation."

I'd say prefixing is desired only in transformers and providers and in debug level. In normal operations, in the rest of the app is not desired - almost no CLI apps do it. 

Going with simple logger could be the way. E.g. having a logging wrapper(s) with the prefixing in transformer and provider base class and using them in the logging calls in transformer/provider implementations.

Tiboris:
From my point of view it is totally all-right to complicate logger little bit so it fulfil our needs and we use the same everywhere and as implementation looks like now, it is not really that complicated. I believe it is okay that project has its own logger.

src/mrack/transformers/transformer.py
```diff
@@ -41,9 +42,13 @@ async def init(self, cfg, metadata):
         self._provider = providers.get(self._config_key)
         await self.init_provider()
 
+    def set_display_name(self, name):
```
netoarmando, line 45: It's a public property, I think this is not necessary,

src/mrack/providers/aws.py
```diff
@@ -76,17 +76,22 @@ async def validate_hosts(self, hosts):
             req_img = req.get("image")
             if not req.get("meta_image") and req_img not in self.ami_ids:
                 raise ValidationError(
-                    f"{self.display_name} provider does not support "
-                    f"'{req_img}' image in provisioning config."
+                    f"{self.display_name}: Provider does not support "
```
netoarmando, line 79: `self.dsp_name`

Tiboris: right!

src/mrack/transformers/transformer.py
Tiboris: allright

src/mrack/logging.py
Tiboris: This is not relevant anymore marking as resolved

netoarmando (approved):

PR #46, opened 2020-11-04
Require specific versions for mrack

Tiboris:
It is always better to require specific versions
of dependencies because pip always installs
the latest and they may break funcionality
or create conflicts when mrack is a dependency.


netoarmando (approved):
Approving this so we can do a release to fix production env.

Versions are not far from latest (some are actually the latest).

PR #48, opened 2020-11-05
Fix regression in up action with OpenStack provider

pvoborni:
Provisioning failed because requirements contained unexpected information.

I.e., the provider should receive only the information it expects or
are valid for OpenStack provisioning.

Removing the culprit lines: meta_image and restraint_id params.


Tiboris (approved):
Thanks for the fixed we have missed this!

PR #49, opened 2020-11-05
Remove unused values in aws and beaker transformation

Tiboris:
These values are loaded from metadata when generating
inventory so they do not need to be passed to provider.


pvoborni:
ACK, even though I did not test in prod. Both Beaker and AWS providers don't used this variable and it is not relevant for provisioning - it is used later by restraint test harness for beaker.

pvoborni (approved):

Tiboris:
I will take responsibility.

PR #50, opened 2020-11-06
Increase OpenStack timeout to 6mins

pvoborni:
Trying to make the provider more stable if OpenStack is slow, e.g. under load.

It's not very clever method. But might work.


Tiboris (approved):
I will approve this :) thanks for the patch.

PR #52, opened 2020-11-06
Do not set username for pytest multihost config

pvoborni:
We have actually not used it at it cannot be assumed that
the user used for provisioning should be used for pytest-multihost.

E.g. a lot of tests rely on root - to be able to completely control
the remote machine. That cannot be done, e.g. with default cloud-user.

(Though it can often sudo to root)

Let's keep there the more simple configuration, where user is the
same for all host and is defined in provisioning config 'mhcfg' dir.


Tiboris (approved):
Hello @pvoborni 

could you please squash the commits?

pvoborni:
Is there a  reason for squashing them? To a certain degree they could be a separate PRs.

Tiboris:
Nevermind.

PR #53, opened 2020-11-06
Fix ssh action and improve its error UX

pvoborni:
Fix OpenStack provisioning status translation
    
    It is not needed to use a STATUS_MAP for translation status in
    prov_result_to_host_data as it is done also in to_host of parent
    class. Double translation doesn't work and thus everything provisioned
    had incorrecly status : "other".
    
    This has then a negative effect on ssh actione where no host is reported
    as active.
    
Do not show traceback on ApplicationError
    
    So that, e.g., on ssh action, when there are no active host, user
    will not get an exception ending with:
    
    ```
    raise ApplicationError("No active host available.")
    ```
    
    But only:
    ```
    No active host available.
    ```
    
    Which is more user friendly.
    
### DIFF
--- a/src/mrack/providers/openstack.py
+++ b/src/mrack/providers/openstack.py
@@ -28,7 +28,6 @@ from mrack.host import (
     STATUS_ACTIVE,
     STATUS_DELETED,
     STATUS_ERROR,
-    STATUS_OTHER,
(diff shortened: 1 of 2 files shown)


Tiboris (approved):
Changes work good form me @pvoborni

PR #55, opened 2020-11-11
MrackConfig: improve the approach to be object based and easy to extend 

pvoborni:
base on PR: https://github.com/neoave/mrack/pull/36

This commit implements a MrackConfig which allows to define options
and arguments otherwise needed in a configuration file and thus
save people's time with writing them in CLI.

Config file looks like:
```
[mrack]
mrackdb = config/mrackdb.yaml
provisioning-config = config/provisioning-config.yaml
metadata = config/metadata.yaml
```

Where each value is optional and existance of the config is also
optional.

Mrack looks for the config on these locations:
- user provided path via -c or --mrack-config global option
- ./mrack.conf
- ~/.mrack/mrack.conf
- /etc/mrack/mrack.conf

It also changes CLI of commands:
- metadata is no longer an argument but an option: -m or --metadata
- provisioning config shortcut changes to global -p

With these changes it is possible to define a config file to allow
just typing:
```
$ mrack up
$ mrack list
$ mrack ssh
$ mrack output
$ mrack destroy
```

The MrackConfig class is also extensible for new use cases.

### DIFF
--- a/src/mrack/data/mrack.conf
+++ b/src/mrack/data/mrack.conf
@@ -1,4 +1,4 @@
 [mrack]
 mrackdb = .mrackdb.json
 provisioning-config = provisioning-config.yaml
-
+metadata = metadata.yaml
--- a/src/mrack/data/mrack.conf
+++ b/src/mrack/data/mrack.conf
@@ -1,0 +1,4 @@
+[mrack]
+mrackdb = .mrackdb.json
+provisioning-config = provisioning-config.yaml
+
--- a/.gitignore
+++ b/.gitignore
@@ -1,19 +1,138 @@
-# python
-*.pyc
-*.pyo
-*.swp
-__pycache__
+# Byte-compiled / optimized / DLL files
+__pycache__/
+*.py[cod]
+*$py.class
 
-# packaging folders
-/build/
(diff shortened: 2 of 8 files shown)


src/mrack/data/mrack.conf
```diff
@@ -0,0 +1,4 @@
+[mrack]
+mrackdb = .mrackdb.json
+provisioning-config = provisioning-config.yaml
+
```
Tiboris, line 4: Could you please add here `metadata` key value example?

Tiboris: > Mrack looks for the config on these locations:
>
>    user provided path via -c or --mrack-config global option
>    ~./mrack.conf

I believe `~` should not be there

pvoborni: Commit message ammended.

pvoborni: And metadata added to data/mrack.conf

Tiboris:
Works for me @pvoborni thanks for the patch and improvement :)

Tiboris (approved):
Works for me @pvoborni thanks for the patch and improvement :)

PR #57, opened 2020-11-17
Simple way to add hosts into /etc/hosts file

pvoborni:
This is useful for cases where the fake hostname needs to be used
e.g. when accessing FreeIPA Web UI which is sensitive on used hostname.

It is uncomfortable task to read the values from `mrack list` and add
them manualy to /etc/hosts.

The change introduces two commands to help with this case:

* `$ mrack eh add` to add all active hosts into /etc/hosts
* `$ mrack eh clear` to remove all records added by mrack.

Mrack adds two markers into the /etc/hosts file and adds records
between them. This should work also with multiple runs of mrack in
separate tests. If the section already contains the fake hostname then
this line is overwritten with a new value. This is useful when
provisioning hosts for the same task after they were previously
removed.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,3 +4,8 @@ include_trailing_comma = true
 known_first_party = "mrack"
 line_length = 88
 multi_line_output = 3
+
+[tool.pytest.ini_options]
+markers = [
+    "etc_hosts: passes path for etc_hosts fixture",
+]
--- a/src/mrack/actions/etchosts.py
+++ b/src/mrack/actions/etchosts.py
@@ -1,0 +1,146 @@
+# Copyright 2020 Red Hat Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Etc-Hosts actions."""
+
+import getpass
+import logging
+
+from mrack.errors import ApplicationError, ConfigError
+from mrack.host import STATUS_ACTIVE
+
+logger = logging.getLogger(__name__)
+
+MRACK_START = "# Managed by mrack - start\n"
+MRACK_END = "# Managed by mrack - end\n"
+
+
+class EtcHostsUpdater:
+    """Object to manipulate with /etc/hosts file."""
+
+    def __init__(self, path="/etc/hosts"):
+        """Init the updater."""
+        self.path = path
+        with open(self.path, "r") as f:
+            self.lines = f.readlines()
+
+        # Trigger validation
+        self._locate_mrack_section()
+
+    def _locate_mrack_section(self):
+        """Locate mrack managed section."""
+        start = -1
+        end = -1
+        mrack_lines = []
+        errors = []
+        for index, line in enumerate(self.lines):
+            if MRACK_START in line:
+                if start != -1:
+                    errors.append("Multiple start marks.")
+                start = index
+            if MRACK_END in line:
+                if end != -1:
+                    errors.append("Multiple end marks.")
+                end = index
+            if start != -1 and end == -1 and index > start:
+                mrack_lines.append([index, line])
+
+        # Check errors:
+        if start == -1 and end != -1:
+            errors.append("Doesn't have start mark.")
+        if start != -1 and end == -1:
+            errors.append("Doesn't have end mark.")
+        if end < start:
+            errors.append("End mark before start mark.")
+
+        if errors:
+            error = "/etc/hosts has broken mrack managed section and needs fixing:\n"
+            for err in errors:
+                error += f"   {err}\n"
+            raise ConfigError(error)
+
+        return (start, end, mrack_lines)
+
+    def add_hosts(self, hosts):
+        """Add lines for hosts."""
+        start, end, mrack_lines = self._locate_mrack_section()
+
+        if start == -1:
+            start = len(self.lines)
+            self.lines.append(MRACK_START)
+            end = start + 1
+            self.lines.append(MRACK_END)
+
+        for host in hosts:
+            new_line = f"{host.ip} {host.name}\n"
+            for mrack_line in mrack_lines:
+                idx, line = mrack_line
+                # update: /etc/hosts has already a line for the host
+                if host.name in line:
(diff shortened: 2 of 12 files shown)


tests/unit/utils.py
```diff
@@ -11,3 +12,18 @@ def get_data(name, data_dir="data"):
...
+
+
+def copy_as_temp(data_path, text=True):
+    """
+    Makes a temporary copy of file relative to test dir and return it file path.
+    """
+    source_path = os.path.join(test_dir_path(), data_path)
+    temp_fp, temp_path = tempfile.mkstemp(text=text)
+
+    with open(temp_fp, "w") as out_f:
```
netoarmando, line 24: Possible tip: I *think* you can use [`shutil.copy`](https://docs.python.org/3.8/library/shutil.html#shutil.copy), no need to open the file.

pyproject.toml
```diff
@@ -4,3 +4,8 @@ include_trailing_comma = true
 known_first_party = "mrack"
 line_length = 88
 multi_line_output = 3
+
+[tool.pytest.ini_options]
+markers = [
+    "etc_hosts: passes path for etc_hosts fixture",
+]
```
netoarmando, line 11: Newline at the end.

src/mrack/actions/etchosts.py
```diff
@@ -0,0 +1,148 @@
...
+    def update(self):
+        """Update /etc/hosts."""
+        logger.info("Adding hosts to /etc/hosts file")
+
+        hosts = [h for h in self._db_driver.hosts.values() if h.status == STATUS_ACTIVE]
+        self.updater.add_hosts(hosts)
+        self.updater.save()
+
+        logger.info("Done")
+        return True
```
netoarmando, line 138: Are you planning to check the returned boolean from these methods?

src/mrack/run.py
```diff
@@ -196,6 +197,29 @@ async def ssh(ctx, hostname, metadata):
     ssh_action.ssh(hostname)
 
 
+@mrackcli.group()
+async def eh():
```
Tiboris, line 201: The naming could be better, can we use `hosts` for example ?

tests/pytest.ini
```diff
@@ -0,0 +1,3 @@
+[pytest]
+markers =
+    etc_hosts: passes path for etc_hosts fixture
```
Tiboris, line 3: Here is newline missing at the end :(

Tiboris (commented):
Thansk for the patch @pvoborni!

Is it possible that you could divide this work into two commits for feature and tests?

tests/unit/utils.py
pvoborni: Fixed. Funny thing, wanted to use it originally, but then I though that `mkstemp` will return an open file. It doesn't, I fixed it by adding open instead of returning back.

pyproject.toml
pvoborni: Fixed.

src/mrack/actions/etchosts.py
pvoborni: No reason. A copy&paste issue. Removed.

src/mrack/run.py
pvoborni: I considered "hosts" but find it misleading as mrack's main functionality is about provisioning hosts and thus people might expect to work with the hosts somehow.  Then something like `etc-hosts` is too long and difficult to type. Thus ended with `eh` which is easy to remember and fast to type.

tests/pytest.ini
pvoborni: Fixed.

pvoborni:
> Is it possible that you could divide this work into two commits for feature and tests?

I don't see an added value of the task to spend time on it. Tests are part of the change and this is not e.g. FreeIPA where there were are valid reasons - easier backporting of only tests.

Tiboris:
We need rebase here.

tests/pytest.ini
netoarmando, line 3: I had to add this to root's `pytest.ini` to remove the warnings.

tests/unit/test_etchosts.py
```diff
@@ -0,0 +1,147 @@
...
+
+import os
+
+import pytest
+
+from mrack.actions.etchosts import EtcHostsUpdater
+from mrack.errors import ConfigError
+from mrack.host import Host
+
+from .utils import copy_as_temp
```
netoarmando, line 23: Do you think it's okay to use `from unit.utils import copy_as_temp`?

src/mrack/run.py
Tiboris: What about `patch-hosts` i do not mind at all that it is little bit longer when it says the purpose.

tests/unit/test_etchosts.py
pvoborni: Interesting, it actually works. Changed. We might want to fix (in a different) patch a similar thing in the other test.

tests/pytest.ini
pvoborni: Moved, the other removed.

src/mrack/run.py
pvoborni: `patch-hosts` is the same as `etc-hosts`, maybe actually worse as is longer. What could be OK, is to have a longer alias, i.e. both `eh` and `etc-hosts`. But really, don̈́t see a need, current help is quite self-explanatory.

```
$ mrack
Usage: mrack [OPTIONS] COMMAND [ARGS]...

  Multihost human friendly provisioner.

Options:
  -c, --mrack-config PATH
  -p, --provisioning-config PATH
  -d, --db PATH
  --debug
  --help                          Show this message and exit.

Commands:
  destroy  Destroy provisioned hosts.
  eh       Commands to update /etc/hosts file.
  list     List host tracked by.
  output   Create outputs - such as Ansible inventory.
  ssh      SSH to host.
  up       Provision hosts.


$ mrack eh
Usage: mrack eh [OPTIONS] COMMAND [ARGS]...

  Commands to update /etc/hosts file.

Options:
  --help  Show this message and exit.

Commands:
  add    Add active hosts to /etc/hosts file.
  clear  Remove all mrack hosts from /etc/hosts file.
```

netoarmando (approved):
LGTM

src/mrack/actions/etchosts.py
```diff
@@ -0,0 +1,146 @@
...
+        try:
+            with open(self.path, "w") as f:
+                f.writelines(self.lines)
+        except PermissionError:
+            name = getpass.getuser()
+            err = (
+                f"Error: Not enough permissions to write into file: {self.path}.\n\n"
+                "Consider running mrack as root for this command e.g. via sudo "
+                "or add yourself the necessary rights, e.g. by: \n"
+                f"  $ sudo setfacl -m {name}:rw {self.path}"
```
netoarmando, line 116: Great UX!

Tiboris (approved):
@pvoborni Thanks for the patch!

PR #58, opened 2020-11-17
feat: Output file paths configurable in mrack config

pvoborni:
Adding a possibility to change Ansible inventory and pytest multihost
configuration output file paths in a mrack config file. So that user
can use directly a custom location.

Example of such config:

```ini
[mrack]
mrackdb = config/mrackdb.json
provisioning-config = config/provisioning-config.yaml
metadata = config/metadata.yaml
ansible-inventory = config/test.inventory.yaml
pytest-multihost = config/pytest-multihost-config.yaml
```

### DIFF
--- a/src/mrack/outputs/ansible_inventory.py
+++ b/src/mrack/outputs/ansible_inventory.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 """Ansible inventory output module."""
+import logging
 import typing
 from copy import deepcopy
 
@@ -31,6 +32,8 @@ DEFAULT_INVENTORY_PATH = "mrack-inventory.yaml"
 DEFAULT_INVENTORY_LAYOUT: typing.Dict[str, typing.Dict] = {
     "all": {"children": {}, "hosts": {}}
 }
+
+logger = logging.getLogger(__name__)
 
 
 def copy_meta_attrs(host, meta_host, attrs):
@@ -106,12 +109,12 @@ class AnsibleInventoryOutput:
     files and information in DB.
     """
 
-    def __init__(self, config, db, metadata, path=DEFAULT_INVENTORY_PATH):
+    def __init__(self, config, db, metadata, path=None):
         """Init the output module."""
         self._config = config
         self._db = db
         self._metadata = metadata
-        self._path = path
+        self._path = path or DEFAULT_INVENTORY_PATH
 
     def create_ansible_host(self, name):
         """Create host entry for Ansible inventory."""
@@ -206,4 +209,6 @@ class AnsibleInventoryOutput:
         """Create the target output file."""
         inventory = self.create_inventory()
         save_yaml(self._path, inventory)
+        if self._path:
+            logger.info(f"Created: {self._path}")
         return inventory
--- a/src/mrack/actions/output.py
+++ b/src/mrack/actions/output.py
@@ -29,11 +29,15 @@ class Output:
     Test action to run output modules from data in database.
     """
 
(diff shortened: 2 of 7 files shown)


Tiboris:
This needs rebase please, and aware of conflicts.

src/mrack/outputs/ansible_inventory.py
```diff
@@ -106,12 +109,12 @@ class AnsibleInventoryOutput:
     files and information in DB.
     """
 
-    def __init__(self, config, db, metadata, path=DEFAULT_INVENTORY_PATH):
+    def __init__(self, config, db, metadata, path):
```
netoarmando, line 112: I *think* you need to pass `path=None`, so `DEFAULT_INVENTORY_PATH` can be used. (I didn't test it)

tests/unit/test_ansible_inventory.py
```diff
@@ -89,7 +89,7 @@ class TestAnsibleInventory:
     def test_layouts(self, layout, db, metadata):
 
         config = provisioning_config(layout)
-        ans_inv = AnsibleInventoryOutput(config, db, metadata)
+        ans_inv = AnsibleInventoryOutput(config, db, metadata, None)
```
netoarmando, line 92: Relevant comment: https://github.com/neoave/mrack/pull/58/files#r529522230
nit-pick: personally I don't like to see `None` passed as positional arguments, doesn't look good.

Tiboris:
```suggestion
        ans_inv = AnsibleInventoryOutput(config, db, metadata, path=None)
```

src/mrack/outputs/ansible_inventory.py
```diff
@@ -206,4 +209,6 @@ def create_output(self):
         """Create the target output file."""
         inventory = self.create_inventory()
         save_yaml(self._path, inventory)
+        if self._path:
+            logger.info(f"Created: {self._path}")
```
Tiboris, line 213: In these, could you please use lazy formatting?
`W1203: Use lazy % formatting in logging functions (logging-fstring-interpolation)`
I future pylint will not complain.

Tiboris (changes requested):
Thanks for the PR @pvoborni for the latest changes please use suggestions, otherwise it looks good to me.

If the commit-checker action will fail after force push, please update PR's desctiption to fulfil the needs (prepend it with the feat: Output file paths...)

src/mrack/outputs/ansible_inventory.py
pvoborni: Is it actually needed? By rough estimates, 2/3 of logger.info calls in the code base are using this method. I believe there is none with lazy formatting. Atm. it will make it more inconsistent with rest of code base.

Also, I believe that with W1203 we would be imposing unecessary purity over readability.  W1203 is about performance and performance impact with f-strings is almost 0.

Tiboris: Let us mark it as resolved then

Tiboris (approved):
Thanks @pvoborni let us merge it.

src/mrack/outputs/ansible_inventory.py
netoarmando: Personally, I have no opinion about this, but recently I was faced with the same dilemma on another project and found this: https://docs.python.org/3/howto/logging-cookbook.html#formatting-styles

One _possible_ reason for using `%`: mrack can be imported as a module and the main project can independently decide how logs are handled (Assuming backwards compatibility).

PR #59, opened 2020-11-18
Add pre-commit configuration files

Tiboris:
Add pre-commit configuration files and make changes to sources so they pass the checks

Updating the files that yamllint was complaining about

Fixing the pre-commit eof-check failures

Remove deprecated flag --recursive from Makefile for the isort
Add command to sort setup.py.
Update .isort.cfg to be compatible with black.

Update files after make format
Previous commit which updated .isort.cfg file
would cause failures in testing with black
and isort afterwards.


Makefile
```diff
@@ -1,9 +1,10 @@
 format:
-	isort --recursive src
```
pvoborni, line 2: I guess I'll finally upgrade to Fedora 33 from F31, where there is no isort 5.  :)

README.md
```diff
@@ -95,15 +95,18 @@ import mrack
...
+Projects is using [black](https://github.com/psf/black) formater and [isort](https://github.com/PyCQA/isort) to keep consistent
 formatting, [flake8](https://flake8.pycqa.org/en/latest/) and
 [pydocstyle](http://pycodestyle.pycqa.org/en/latest/intro.html) to ensure following
 Python good practices.
 
-Contributions (Pull Requests) are welcome. It is expected that they will pass tox tests
-and include unit tests for new code.
+Contributions (Pull Requests) are welcome. It is expected that they will pass tox tests and code checkers.
+Inclusion of the unit tests for the new code is recommended.
+Because of that we have configured [pre-commit](https://github.com/pre-commit/pre-commit) hook.
```
pvoborni, line 105: Isn't it better to point to https://pre-commit.com/ , i.e. one less click?

.pre-commit-config.yaml
```diff
@@ -0,0 +1,53 @@
...
+  - repo: https://github.com/ambv/black
+    rev: 20.8b1
+    hooks:
+      - id: black
+        exclude: ^docs/.*
+        types: [python]
+        args: [--safe, --quiet, --line-length, "88"]
+        require_serial: true
+
+  - repo: https://github.com/asottile/seed-isort-config
```
pvoborni: Should it be here? This PR is requiring us to use isort 5 (given the removal of --recursive).  The repo is archived and on the web page https://github.com/asottile-archive/seed-isort-config  they say: "this is no longer needed as of isort>=5".

pvoborni (changes requested):
Thanks, overall I like it but there are couple things to fix. 

I guess that it will also need to be added to Azure pipelines config, otherwise pre-commit won't be run in gating.

.pre-commit-config.yaml
Tiboris: Removed

Makefile
Tiboris: ohhh :D sorry maybe its time as the 31 is already "stable"

azure-pipelines.yml
```diff
@@ -1,69 +1,70 @@
...
-- master
+trigger: master
 
 pool:
   vmImage: 'ubuntu-latest'
 strategy:
   matrix:
     Python37:
-      python.version: '3.7'
+      python.version: '3.9'
```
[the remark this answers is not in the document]
Tiboris: will try to fix

pvoborni (approved):

README.md
Tiboris: Let me change it to the official site.

netoarmando (approved):

PR #63, opened 2020-11-24
chore: Add release workflow to github actions

Tiboris:
Added release workflow to be run by hand:
workflow-image
This workflow depends on test and pre-commit runs:
dep-image
And after these succeeds runs the python-semantic-release action:
release-image
In this action:
- a new commit with changed version and changelog is pushed to master
- a new tag is added to the commit above with particular release version
- a new mrack package is being release to PyPi (prerequisite: secret with PYPI_TOKEN - already set)

Please note: the release 0.3.1 will not be created becaue it has been already this is why test action is failing here: https://github.com/Tiboris/mrack/runs/1448905490?check_suite_focus=true (see: https://pypi.org/help/#file-name-reuse)

### DIFF
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -13,5 +13,5 @@ jobs:
       - uses: actions/checkout@v2
       - uses: actions/setup-python@v2
         with:
-          python-version: '3.9'
+          python-version: 3.9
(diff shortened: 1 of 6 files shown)


netoarmando (approved):
LGTM

pvoborni (approved):
LGTM

PR #64, opened 2020-11-25
chore: Add pylint to pre-commit-config.yaml and fix its issues in mrack

Tiboris:
Use pylint in pre-commit-hook and actions as well to check code for easy to overlook mistakes.

Fix pylint isssues in mrack project

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -53,3 +53,12 @@ repos:
       - id: mypy
         exclude: ^(docs|tests)/.*$
         args: ["--ignore-missing-imports"]
+
+  - repo: https://github.com/pre-commit/mirrors-pylint
+    rev: v2.6.0
+    hooks:
+      - id: pylint
+        additional_dependencies: ["isort[pyproject]"]
+        exclude: ^(docs/|tests/).*$
+        # disabled import-error as may be run out of environment with deps
+        args: ["--disable=import-error"]
--- a/src/mrack/host.py
+++ b/src/mrack/host.py
@@ -41,9 +41,9 @@ def host_from_json(host_data):
     provider = providers.get(provider_name)
     host = Host(
         provider,
-        host_data["id"],
+        host_data["host_id"],
         host_data["name"],
-        host_data["ips"],
+        host_data["ip_addrs"],
         host_data["status"],
         host_data["rawdata"],
         host_data["username"],
@@ -62,9 +62,9 @@ class Host:
     def __init__(
         self,
         provider,
-        id,
+        host_id,
         name,
-        ips,
+        ip_addrs,
         status,
         rawdata,
         username=None,
@@ -73,9 +73,9 @@ class Host:
     ):
         """Initialize host object."""
         self._provider = provider
-        self._id = id
+        self._host_id = host_id
         self._name = name
-        self._ips = ips
+        self._ip_addrs = ip_addrs
         self._status = status
         self._username = username
         self._password = password
@@ -84,25 +84,24 @@ class Host:
 
     def __str__(self):
         """Return string representation of host."""
-        net_str = " ".join(self._ips)
+        net_str = " ".join(self._ip_addrs)
 
         out = (
-            f"{self._status} {self._id} {self._name} {net_str} {self._username} "
+            f"{self._status} {self._host_id} {self._name} {net_str} {self._username} "
             f"{self._password}"
         )
 
         if self._error:
-            o = [out, "Error:", object2json(self._error)]
-            out = "\n".join(o)
+            out = "\n".join([out, "Error:", object2json(self._error)])
         return out
 
     def to_json(self):
         """Transform object into representation which is acceptable by `json.dump`."""
         return {
             "provider": self._provider.name,
-            "id": self._id,
+            "host_id": self._host_id,
             "name": self._name,
-            "ips": self._ips,
+            "ip_addrs": self._ip_addrs,
             "status": self._status,
             "username": self._username,
             "password": self._password,
@@ -116,9 +115,9 @@ class Host:
         return self._provider
 
     @property
-    def id(self):
+    def host_id(self):
         """Get provider host id."""
-        return self._id
+        return self._host_id
 
     @property
     def name(self):
@@ -126,15 +125,14 @@ class Host:
         return self._name
 
     @property
-    def ips(self):
+    def ip_addrs(self):
         """Get host IP addresses."""
-        return self._ips
+        return self._ip_addrs
 
     @property
-    def ip(self):
+    def ip_addr(self):
         """Get first host IP address."""
-        if self._ips:
-            return self._ips[0]
+        return self._ip_addrs[0] if self._ip_addrs else ""
 
     @property
     def status(self):
@@ -158,6 +156,6 @@ class Host:
 
     async def delete(self):
         """Issue host deletion via associated provider."""
-        await self.provider.delete_host(self.id)
+        await self.provider.delete_host(self.host_id)
         self._status = STATUS_DELETED
         return True
--- a/src/mrack/providers/aws.py
+++ b/src/mrack/providers/aws.py
@@ -43,7 +43,7 @@ class AWSProvider(Provider):
         self.ssh_key = None
         self.sec_group = None
         self.instance_tags = None
-        self.STATUS_MAP = {
+        self.status_map = {
             "running": STATUS_ACTIVE,
             "pending": STATUS_PROVISIONING,
             "terminated": STATUS_DELETED,
@@ -76,9 +76,9 @@ class AWSProvider(Provider):
         """Prepare provisioning."""
         pass
 
-    async def validate_hosts(self, hosts):
+    async def validate_hosts(self, reqs):
         """Validate that host requirements are well specified."""
-        for req in hosts:
+        for req in reqs:
             req_img = req.get("image")
             if not req.get("meta_image") and req_img not in self.ami_ids:
                 raise ValidationError(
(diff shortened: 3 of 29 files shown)


src/mrack/host.py
```diff
@@ -84,25 +84,24 @@ def __init__(
...
-            out = "\n".join(o)
+            out = "\n".join([out, "Error:", object2json(self._error)])
         return out
 
     def to_json(self):
         """Transform object into representation which is acceptable by `json.dump`."""
         return {
             "provider": self._provider.name,
-            "id": self._id,
+            "host_id": self._host_id,
```
pvoborni, line 102: This is backwards incompatible change. Are we OK with it?

src/mrack/host.py
```diff
@@ -116,25 +115,24 @@ def provider(self):
...
+        return self._host_id
 
     @property
     def name(self):
         """Get host name."""
         return self._name
 
     @property
-    def ips(self):
+    def ip_addrs(self):
```
pvoborni, line 128: Just a rant: This rule, I'm assume minimal number of chars in variable/attribute name, is IMHO stupid - e.g. in cases like this.

src/mrack/providers/aws.py
```diff
@@ -93,13 +93,15 @@ async def validate_hosts(self, hosts):
...
+                raise ValidationError(
+                    f"{err_msg} Request failed with {err_resp}"
+                ) from image_err
 
         return
 
     async def can_provision(self, hosts):
         """Check that hosts can be provisioned."""
-        return True
+        return bool(hosts)
```
pvoborni, line 104: This is bad. `can_provision([])` should return `True`.

src/mrack/run.py
```diff
@@ -256,15 +256,15 @@ def handle(*args, **kwargs):
...
+        return ret_code
 
     return handle
 
 
 @exception_handler
 def run():
     """Run the app."""
-    mrackcli(obj={})
+    mrackcli()  # pylint: disable=no-value-for-parameter
```
pvoborni, line 267: Will this work?  AFAIK the `obj={}` is needed for correct work of context.

pvoborni (dismissed):
Hi thanks for the patch. Mostly OK. Added some early-comments.  

But I must note that I'm not really fan of the "short variable names" rule. In some cases it helps but in a lot it doesn't really add value and makes things longer.

src/mrack/host.py
Tiboris: @netoarmando what do you say

src/mrack/providers/openstack.py
```diff
@@ -268,30 +268,31 @@ def _translate_networks(self, req, spec=False):
...
-        self._translate_image(req)
-        self._translate_networks(req)
+        self._translate_flavor(reqs)
+        self._translate_image(reqs)
+        self._translate_networks(reqs)
 
         return True
 
-    async def validate_hosts(self, reqs):
+    async def validate_hosts(self, hosts):
```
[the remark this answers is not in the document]
Tiboris: Allright we can do so

src/mrack/providers/aws.py
Tiboris: Is this really needed or expected? I think that provider without hosts should not even 'think' about provisioning.

src/mrack/run.py
Tiboris: mrack up and destroy works fine for me without the `obj={}`

src/mrack/providers/openstack.py
```diff
@@ -268,30 +268,31 @@ def _translate_networks(self, req, spec=False):
...
-        for req in reqs:
-            self.validate_host(req)
+        for host in hosts:
+            self.validate_host(host)
 
     def get_host_requirements(self, req):
         """Get vCPU and memory requirements for host requirement."""
         flavor_spec = req.get("flavor")
         flavor_ref = req.get("flavorRef")
+        flavor = self.flavors["default"]
```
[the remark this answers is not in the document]
Tiboris: Fixed

src/mrack/providers/aws.py
pvoborni: It's an edge case. "Can provision" is about resources and it would be weird to say no with 0 hosts and yes with e.g. 1.  To "think" about provisioning is  on somebody else.

Tiboris: I believe that empty requirement is a corner case. Is my assumption correct if I say that this may happen only while using mrack as library.
AWS, Beaker, and Static for example they will always provision (well if Static provider got a empty list instead of existing resource it is a very bad) because there is different approach in openstack we have quota and in these we are 'billed' by instances. I believe this is not a big issue that we return false if this list is empty. 
I believe adding a `req` to the line with `raise ValidationError(f"{self.dsp_name}: Not enough resources to provision")` and having errors more distinguished might be good as well. For example:
```python
        if not await self.can_provision(hosts):
            if not hosts:
                raise ProvisioningError("Can not continue with empty requirement for provider")
            else:
                raise ValidationError(f"{self.dsp_name}: Not enough resources to provision")
        logger.info(f"{self.dsp_name}: Resource availability: OK")
```

netoarmando: Short answer: I think `can_provision` should return True for empty lists, given that there's no other logic there. 

Regarding the `if not hosts -> raise Exception`: this is the kind of validation I would expect to find within that method, but empty list of hosts could print a warning message perhaps. This way mrack as library can handle this before calling `can_provision` and mrack as tool will "never" encounter this.

src/mrack/host.py
netoarmando: I'm okay with this, I like to use verbose variables for things like `id`, `type`, `instance`. 

But based on another comment from Petr, the issue here is the number of characters, right? So, do we need to change the key as well? Would that make this backwards compatible?

src/mrack/run.py
netoarmando: Shouldn't it be `ctx={}` to have some effect?

_(Not actually related interesting "known gotcha": https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments.)_

pvoborni: I vaguely remember that it was needed. It is documented here: https://click.palletsprojects.com/en/7.x/commands/#nested-handling-and-contexts But if **all** commands with various options used work the same without it then we can remove it.

src/mrack/host.py
pvoborni: Issue is that loading/saving mrack db which was created before this change will not work. But I guess it is acceptable given that we did not release a 1.0 yet.

src/mrack/run.py
Tiboris: i reverted this change

src/mrack/providers/aws.py
Tiboris: Fixed

.pre-commit-config.yaml
```diff
@@ -53,3 +53,12 @@ repos:
       - id: mypy
         exclude: ^(docs|tests)/.*$
         args: ["--ignore-missing-imports"]
+
+  - repo: https://github.com/pre-commit/mirrors-pylint
+    rev: v2.6.0
+    hooks:
+      - id: pylint
```
netoarmando, line 60: What's the difference between using this and adding pylint as a command in `tox.ini`? Because this should be part of gating too.

netoarmando (commented):
Regarding `Reusing existing methods in static and podman provider`: Looks good to me. Also, I've tested the static provider as a sanity check.

.pre-commit-config.yaml
Tiboris: There is no difference it is redundant check.

src/mrack/providers/openstack.py
```diff
@@ -432,26 +447,26 @@ async def wait_till_provisioned(
 
         start = datetime.now()
         timeout_time = start + timedelta(minutes=timeout)
-        done_states = ["ACTIVE", "ERROR"]
 
         # do not check the state immediately, it will take some time
         await asyncio.sleep(poll_sleep_initial)
 
-        error_attempts = 0
```
netoarmando, line 440: This variable must be defined, it's used later inside the `while` block.

netoarmando (changes requested):
I think commit `fix: Fix pylint isssues and reuse existing methods` could be split into 2 commits.
In general I was okay with all changes pylint was complaining; I like to use "longer" variables, but I understand we will need to see `# pylint: disable=invalid-name` sometimes.

However, that commit is also refactoring on some methods, I'd like to see that in an additional commit at least.
I left one in-line comment about that refactoring.

src/mrack/providers/openstack.py
Tiboris: my bad i moved that to the argument :D damn

Tiboris:
>However, that commit is also refactoring on some methods, I'd like to see that in an additional commit at least.
I left one in-line comment about that refactoring.

I would really like to but pylint is complaining about lot of stuff and to silence it i did all the changes requested which lead to refactor. Please @netoarmando  could you re check and tell if the changes are acceptable?

Tiboris:
AWS and OpenStack provisioning works with these changes I have tested metadata:
```
domains:
- hosts:
  - group: ipaclient
    name: f33.tests-mrack.test
    os: fedora-33
    role: master
  name: tests-mrack.test

```

Tiboris:
With https://github.com/neoave/mrack/pull/87 beaker works as well.

netoarmando (approved):
LGTM, I double-checked some variables and definitions and I did some testing regarding provisioning, no errors encountered.

PR #65, opened 2020-12-02
fix: Remove double status translation in parse_errors

pvoborni:
When rebasing the Podman provider I've noticed this bug. It was failing for Podman as it has no STATUS_MAP but the line is IMHO incorrect for all others. It may be hard to notice as it defaults to STATUS_OTHER. I did not test if it works though.

Description:

parse_errors works with Host objects which already have the normalized status so usage of STATUS_MAP is incorrect.


Tiboris (approved):
Thanks!

PR #66, opened 2020-12-07
fix: AWS Error for non authorized user reading image + refactor overcomplicated join

Tiboris:
There was an issue when reading image.name for image
for which user did not have proper read rights and
is not authorised to use it - Nonetype has no attribute name

refactor overcomplicated code in openstack in second commit

### DIFF
--- a/src/mrack/providers/aws.py
+++ b/src/mrack/providers/aws.py
@@ -88,6 +88,12 @@ class AWSProvider(Provider):
 
             try:
                 aws_image = self.ec2.Image(req_img)
+                if not aws_image:  # user is not authorized to use ami - None returned
+                    raise ValidationError(
+                        f"{self.dsp_name}: User does not have enough permissions "
+                        f"to use image: {req_img}"
+                    )
+
                 logger.info(
                     f"{self.dsp_name}: Requested provisioning of {aws_image.name} image"
                 )
@@ -98,7 +104,7 @@ class AWSProvider(Provider):
                 )
                 logger.error(err_msg)
                 err_resp = image_err.response["Error"]["Message"]
-                raise ValidationError(f"{err_msg} Request failed with {err_resp}")
+                raise ValidationError(f"{err_msg}. Request failed with {err_resp}")
 
         return
 
--- a/src/mrack/providers/openstack.py
+++ b/src/mrack/providers/openstack.py
@@ -288,11 +288,7 @@ class OpenStackProvider(Provider):
             set([req["image"] for req in reqs if req["image"] not in self.images])
         )
         if prepare_images:
-            if len(prepare_images) > 1:
-                im_list = ", ".join(prepare_images)
-            else:
-                im_list = prepare_images.pop()
-
+            im_list = ", ".join(prepare_images)
             logger.debug(f"{self.dsp_name}: Loading image info for: '{im_list}'")
             await self.load_images(list(prepare_images))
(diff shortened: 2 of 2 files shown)


pvoborni:
To you have a traceback for the original issue? 

Do I understand it correctly that when user doesn't have rights for certain image:
```
aws_image = self.ec2.Image(req_img)
```
Results in `aws_image` being `None` and thus following:
```
logger.info(
      f"{self.dsp_name}: Requested provisioning of {aws_image.name} image"
)
```
Throws attribute error because it is basically: `None.name`. Is it correct?

If so, would it be better to handle it in a following way?
```
aws_image = self.ec2.Image(req_img)
if not aws_image:
    raise ValidationError(
         f"{self.dsp_name}: Used credentials for user who is "
         f"not authorized to use image: {req_img}"
    )
```

pvoborni:
Also what is the error/behaviour if there is no image with the desired name?

Tiboris:
> Results in aws_image being None and thus following:

Yes thats correct this would work
```python
if not aws_image:
    raise ValidationError(
         f"{self.dsp_name}: Used credentials for user who is "
         f"not authorized to use image: {req_img}"
    )
```

> Also what is the error/behaviour if there is no image with the desired name?
```
except ClientError as image_err:
                err_msg = (
                    f"{self.dsp_name}: Requested image "
                    f"'{req_img}' can not be provisioned"
                )
                logger.error(err_msg)
                err_resp = image_err.response["Error"]["Message"]
                raise ValidationError(f"{err_msg} Request failed with {err_resp}")
```

src/mrack/providers/aws.py
```diff
@@ -83,6 +83,12 @@ async def validate_hosts(self, hosts):
 
             try:
                 aws_image = self.ec2.Image(req_img)
+                if not aws_image:  # if user has no rights to use ami None is returned
+                    raise ValidationError(
+                        f"{self.dsp_name}: Used credentials for user who is "
```
pvoborni, line 88: Sorry for not writing it the first time. What about  a message: "Image {req_imd} does not exist or you don't have permissions to read it."? It's shorter, says the same thing + mentions the possibility of the image not existing.

Tiboris: This error is being handled by: ` except ClientError as image_err:`

Tiboris: Or... are you refering to some edge case?

netoarmando: So, this exception(`ValidationError`) will never be raised if image doesn't exist?

Tiboris: Yes that's what I have encountered.

Tiboris: There will be an exception with message that ami is invalid I hope i am not mistaken.

pvoborni: Btw, don't wait for me on finishing the PR. I'm leaving the rest on you as my main point was addressed. 

Though, I'd still recommend to use a simple error message like: `You don't have permissions to use image: {req_img}` I.e. `Used credentials for user ` == `you`

Tiboris:
With https://github.com/neoave/mrack/pull/66/commits/ca82c4604a2e9eefdd2ebbc1884b07384e567f5a i try to fix overcomplicated code I created.

netoarmando (approved):

PR #67, opened 2020-12-08
fix: mrack.conf '~' causes no such file or directory

Tiboris:
When `~` is used in path for mrack.conf the later
operation open('`~/.mrack/mrackdb.json`', 'w') fails with an
exception FileNotFoundError:
[Errno 2] No such file or directory: '~/.mrack/mrackdb.json'

### DIFF
--- a/src/mrack/providers/beaker.py
+++ b/src/mrack/providers/beaker.py
@@ -101,7 +101,7 @@ class BeakerProvider(Provider):
 
     def _allow_ssh_key(self, ssh_key):
 
-        with open(ssh_key, "r") as key_file:
+        with open(os.path.expanduser(ssh_key), "r") as key_file:
             key_content = key_file.read()
 
         return [
--- a/src/mrack/utils.py
+++ b/src/mrack/utils.py
@@ -20,6 +20,7 @@ import contextlib
 import datetime
 import json
 import logging
+import os
 import sys
 
 import yaml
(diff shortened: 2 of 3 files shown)


pvoborni:
Isn't this a problem of the consumer of the value? I.e. the DB should handle it, not config. Why have this "magic" in a general `get` method? But if we want to be really expanding the values in config then it should be done only for the values which are paths.  

Btw, the use case presented in this issue is something I'd not recommend to use. Atm, the mrack DB don't have concept of sessions so the DB itself is one session and thus it is not something for sharing.

Tiboris:
I have moved the expanduser to the open() calls

pvoborni:
Did not test yet, but the approach looks good to me.

Tiboris:
Before:
```
(venv) ➜  venv git:(master) cat mrack.conf 
[mrack]
mrackdb = mrackdb.json
provisioning-config = ~/Work/files/provisioning-config.yaml
(venv) ➜  venv git:(master) file ~/Work/files/provisioning-config.yaml
/home/tdudlak/Work/files/provisioning-config.yaml: ASCII text
(venv) ➜  venv git:(master) mrack up -m ./metadata-f33.yaml 
Provisioning config file not found: ~/Work/idm-ci/files/provisioning-config.yaml
(venv) ➜  venv git:(master) echo $?
1
```
with this patch:
```
(venv) ➜  venv git:(master) git co user-path 
Switched to branch 'user-path'
Your branch is up to date with 'origin/user-path'.
(venv) ➜  venv git:(user-path) cat mrack.conf 
[mrack]
mrackdb = mrackdb.json
provisioning-config = ~/Work/idm-ci/files/provisioning-config.yaml
(venv) ➜  venv git:(user-path) mrack up -m ./metadata-f33.yaml
OpenStack: Initializing provider
OpenStack: Login duration 0:00:03.698010
^C
Aborted!
```

netoarmando (approved):

PR #68, opened 2020-12-09
fix: Fix making ssh_key_filename absolute for default behavior.

pvoborni:
The logic was not working as the comment said. The cause was that
`mhcfg` already contained the value from provisioning config. So
we need to check the metadata to detect the situation.

### DIFF
--- a/src/mrack/outputs/ansible_inventory.py
+++ b/src/mrack/outputs/ansible_inventory.py
@@ -14,6 +14,7 @@
 
 """Ansible inventory output module."""
 import logging
+import os
 import typing
 from copy import deepcopy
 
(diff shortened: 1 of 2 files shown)


pvoborni:
Not tested yet.

Tiboris (commented):
Generation of inventory with this change works like a charm and has abspaths;

I tested also with this change:
```
diff --git a/src/mrack/outputs/ansible_inventory.py b/src/mrack/outputs/ansible_inventory.py
index 74a1cca..6264fe8 100644
--- a/src/mrack/outputs/ansible_inventory.py
+++ b/src/mrack/outputs/ansible_inventory.py
@@ -13,6 +13,8 @@
 # limitations under the License.
 
 """Ansible inventory output module."""
+
+import os
 import logging
 import typing
 from copy import deepcopy
@@ -156,7 +158,7 @@ class AnsibleInventoryOutput:
             host_info["meta_parent_domain"] = meta_domain["parent"]
 
         if ssh_key:
-            host_info["ansible_ssh_private_key_file"] = ssh_key
+            host_info["ansible_ssh_private_key_file"] = os.path.abspath(ssh_key)
 
         if password:
             host_info["ansible_password"] = password
```

Tiboris (approved):
Thanks for the patch! LGTM lets fix this

PR #70, opened 2020-12-09
feat: Retry provisioning strategy

pvoborni:
Optional (currently applied only for OpenStack) provisioning strategy
to retry provisioning of host which ended with error in a provisioning
attempt.

This is useful if environment gets into state where the provisioning
is a bit unstable but in general working. I.e. succeeds for some host
and fails for others.

The Retry strategy:
1. tries to provision all hosts
2. if some failed, delete them and retry only for these hosts
3. if they fail more than max_attempts time(default: 5), fail the job

Default strategy:
- do 1.
- if some failed, fail the job

In each job failure case:
- delete all hosts

### DIFF
--- a/src/mrack/providers/provider.py
+++ b/src/mrack/providers/provider.py
@@ -22,6 +22,9 @@ from mrack.host import STATUS_ERROR, STATUS_OTHER, Host
 
 logger = logging.getLogger(__name__)
 
+STRATEGY_ABORT = "abort"
+STRATEGY_RETRY = "retry"
+
 
 class Provider:
     """General Provider interface."""
@@ -30,6 +33,7 @@ class Provider:
         """Initialize provider."""
         self._name = "dummy"
         self.dsp_name = "Dummy"
+        self.strategy = STRATEGY_ABORT
         self.STATUS_MAP = {"OTHER": STATUS_OTHER}
         return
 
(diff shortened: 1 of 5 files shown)


src/mrack/providers/provider.py
```diff
@@ -106,34 +97,97 @@ async def provision_hosts(self, hosts):
         logger.info(f"{self.dsp_name}: Provisioning duration: {provi_duration}")
 
         hosts = [self.to_host(srv) for srv in server_results]
-        errors = self.parse_errors(hosts)
+        error_hosts = self.parse_error_hosts(hosts)
+        success_hosts = [h for h in hosts if h not in error_hosts]
```
netoarmando, line 101: Instead of several _list comprehension_ calls you can replace them with one `for` loop to fill those lists. This can be considered a nitpick because the number of hosts is usually not big.

netoarmando (approved):
LGTM, I did a quick sanity-check-run, everything went well, but I haven't hit the 'retry' block.

src/mrack/providers/provider.py
```diff
@@ -106,34 +97,97 @@ async def provision_hosts(self, hosts):
...
+            success_hosts.extend(s_hosts)
+
+            if error_hosts:
+                count = len(error_hosts)
+                err = f"{count} hosts were not provisioned properly, deleting."
+                logger.info(f"{self.dsp_name}: {err}")
+                for host in error_hosts:
+                    logger.error(f"{self.dsp_name}: Error: {str(host.error)}")
+                await self.delete_hosts(error_hosts)
+                logger.info("Retrying to provision these hosts.")
```
netoarmando, line 163: `"Retrying to provision these hosts. (Attempt {attempts} of {self.max_attempts}"` that might help us and users.

PR #71, opened 2020-12-14
fix: Set user for Windows host in pytest-multihost config

pvoborni:
Commit 7bb230e170ac0a2373a2316ef23a26bfcb681ad9 removed setting of
users as it did not work "well" for Linux host in our testing. But
it broke assumtion that Windows host have it set. So this is a fix
to be compatible with our previous mechanism.

I'd call it a hotfix as it is needed but not great. It would be
better to have a more general mechanism with a degree of configuration
to make it adjustable for a project.


netoarmando (approved):
I understand the compromise made to fix the issue, LGTM.

PR #75, opened 2020-12-18
chore: change pinned version from strict to range

netoarmando:
While still a good practice to pin the versions of dependencies we found some issues with conflicting dependencies.

This change tries to install greater than or equal to latest major version (or minor if versio is 0.X).

It keeps the exact version (`==`) for boto3 and botocore because of a known conflict in the production environment.


Tiboris (approved):
Lets go for it

PR #76, opened 2020-12-18
fix: mrack eh add no longer complains about coroutine 'eh' not awaited

pvoborni:
Turns:
```
$ mrack eh add
/usr/lib/python3.9/site-packages/click/core.py:1256: RuntimeWarning: coroutine 'eh' was never awaited
  Command.invoke(self, ctx)
RuntimeWarning: Enable tracemalloc to get the object allocation traceback
Adding hosts to /etc/hosts file
Done
```

Into:
```
$ mrack eh add
Adding hosts to /etc/hosts file
Done
```

### DIFF
--- a/src/mrack/run.py
+++ b/src/mrack/run.py
@@ -208,14 +208,13 @@ async def ssh(ctx, hostname, metadata):
 
 
 @mrackcli.group()
-async def eh():
+def eh():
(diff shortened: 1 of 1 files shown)


Tiboris (approved):
Thanks for the patch!

Issue #79, opened 2021-01-13
After OpenStack provisioning times out mrack does not treat this as a failure


Tiboris:
When provisioning with openstack takes more than hour mrack should mar this as failure so machines are returned:
```
2021-01-11T08:57:25 Provision machines
2021-01-11T09:57:35   rc: 1
2021-01-11T09:57:35   start: '2021-01-11 08:57:26.339037'
2021-01-11T09:57:35   stderr: |-
2021-01-11T09:57:35     OpenStack: Initializing provider
2021-01-11T09:57:35     OpenStack: Login duration 0:00:00.212062
2021-01-11T09:57:35     OpenStack: Environment objects load duration: 0:00:01.682436
2021-01-11T09:57:35     Provisioning started
2021-01-11T09:57:35     OpenStack: Created requirement(s): [
2021-01-11T09:57:35         {
2021-01-11T09:57:35             "flavor": "micro",
2021-01-11T09:57:35             "image": "image-latest",
2021-01-11T09:57:35             "key_name": "sshkey",
2021-01-11T09:57:35             "name": "hostname1",
2021-01-11T09:57:35             "network": "IPv4"
2021-01-11T09:57:35         },
2021-01-11T09:57:35         {
2021-01-11T09:57:35             "flavor": "medium",
2021-01-11T09:57:35             "image": "win-2019",
2021-01-11T09:57:35             "key_name": "sshkey",
2021-01-11T09:57:35             "name": "adcveg.domaincveg.com",
2021-01-11T09:57:35             "network": "IPv4"
2021-01-11T09:57:35         }
2021-01-11T09:57:35     ]
2021-01-11T09:57:35     OpenStack: Preparing provider resources
...
2021-01-11T09:57:35     OpenStack: Host fb97bdf8-fdb5-48a2-97a6-509c218d10bb was provisioned in 267.3s
2021-01-11T09:57:35     OpenStack: Host 4d23375a-fe1e-4249-a488-d17da73ae393 was not provisioned within a timeout of 60 mins
2021-01-11T09:57:35     OpenStack: All hosts reached provisioning final state (ACTIVE or ERROR)
2021-01-11T09:57:35     OpenStack: Provisioning duration: 1:00:04.638998
2021-01-11T09:57:35     OpenStack: Printing provisioned hosts
2021-01-11T09:57:35     OpenStack: active fb97bdf8-fdb5-48a2-97a6-509c218d10bb hostname1 192.168.153.56 None None
[...]

PR #80, opened 2021-01-19
fix: treat timeouted machine as an error

Tiboris:
handle TIMEDOUT state and treat it as an error 

Aims to resolve https://github.com/neoave/mrack/issues/79

### DIFF
--- a/src/mrack/providers/provider.py
+++ b/src/mrack/providers/provider.py
@@ -18,7 +18,7 @@ import logging
 from datetime import datetime
 
 from mrack.errors import ProvisioningError, ValidationError
-from mrack.host import STATUS_ERROR, STATUS_OTHER, Host
+from mrack.host import STATUS_ACTIVE, STATUS_OTHER, Host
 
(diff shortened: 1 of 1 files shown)


src/mrack/providers/provider.py
```diff
@@ -181,7 +181,7 @@ def parse_error_hosts(self, hosts):
         logger.debug(f"{self.dsp_name}: Checking provisioned hosts for errors")
         for host in hosts:
             logger.debug(f"{self.dsp_name}: Host - {host.id}\tStatus - {host.status}")
-            if host.status == STATUS_ERROR:
+            if host.status in (STATUS_ERROR, STATUS_TIMED_OUT):
```
netoarmando, line 184: I wonder if we should invert the logic here and check for statuses that we **know** are not considered errors. Just to be "future proof". Does it make sense?

Tiboris: I jump on this right now

netoarmando (approved):
LGTM!

PR #84, opened 2021-02-02
fix: Handle provision errors

netoarmando:
Handling all nova client calls as done in 46048c01f6ae3224e9d49e711e610b28497f63df.

Also, if mrack fails to provision some hosts it must not return 0, an Ansible playbook might misinterpret that as a success.

### DIFF
--- a/src/mrack/providers/openstack.py
+++ b/src/mrack/providers/openstack.py
@@ -26,6 +26,7 @@ from simple_rest_client.exceptions import NotFoundError, ServerError
 
 from mrack.errors import (
     NotAuthenticatedError,
+    ProviderError,
     ProvisioningError,
     ServerNotFoundError,
     ValidationError,
@@ -43,6 +44,7 @@ logger = logging.getLogger(__name__)
 
 PROVISIONER_KEY = "openstack"
 SERVER_ERROR_RETRY = 5
+SERVER_ERROR_SLEEP = 10
 
 
 class OpenStackProvider(Provider):
@@ -420,7 +422,20 @@ class OpenStackProvider(Provider):
         if specs.get("network"):
             del specs["network"]
 
-        response = await self.nova.servers.create(server=specs)
+        error_attempts = 0
+        while True:
+            try:
+                response = await self.nova.servers.create(server=specs)
+            except ServerError as exc:
+                logger.debug(exc)
+                error_attempts += 1
+                if error_attempts > SERVER_ERROR_RETRY:
+                    raise ProvisioningError(
+                        f"{self.dsp_name}: Fail to create server", specs
+                    ) from exc
+                await asyncio.sleep(SERVER_ERROR_SLEEP)
+            else:
+                break
         return response.get("server")
 
     async def delete_server(self, uuid):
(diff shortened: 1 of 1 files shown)


src/mrack/providers/provider.py
```diff
@@ -140,6 +140,10 @@ async def provision_hosts(self, reqs):
         for host in success_hosts:
             logger.info(f"{self.dsp_name}: {host}")
 
+        if error_hosts:
+            # Forcing exit code to be an error
+            raise ProvisioningError("Some hosts couldn't be provisioned")
```
netoarmando, line 145: This will leave successfully provisioned hosts provisioned. We should roll-back and destroy all hosts in this case.

pvoborni: The deletion of hosts should "in theory" happen already in the
```python
       if error_hosts:
            hosts_to_delete = success_hosts + error_hosts
            await self.abort_and_delete(hosts_to_delete, error_hosts)
```
In the method there is:
```python
    async def abort_and_delete(self, hosts_to_delete, error_hosts):
        """Delete hosts and abort provisioning with an error."""
        logger.info(f"{self.dsp_name}: Aborting provisioning due to error.")
        for host in error_hosts:
            logger.error(f"{self.dsp_name}: Error: {str(host.error)}")

        logger.info(f"{self.dsp_name}: Given the error, will delete hosts")
        await self.delete_hosts(hosts_to_delete)
        raise ProvisioningError(error_hosts)
```

So something else might not be working.

Tiboris: Could you please rebase @netoarmando ?

netoarmando: I've re-based it and removed this commit after a discussion with @Tiboris.
We agreed that changes made since last release may have fixed this issue.

I also included a simple sleep between retries.

Tiboris (changes requested):
Hello @netoarmando there is one nitpick from my side please have a look it is nothing serious.

src/mrack/providers/openstack.py
```diff
@@ -420,21 +421,43 @@ async def create_server(self, req):
...
+            try:
+                response = await self.nova.servers.create(server=specs)
+            except ServerError as exc:
+                logger.debug(exc)
+                error_attempts += 1
+                if error_attempts > SERVER_ERROR_RETRY:
+                    raise ProvisioningError(
+                        f"{self.dsp_name}: Fail to create server", specs
+                    ) from exc
+                await asyncio.sleep(1)
```
Tiboris, line 435: Should not we consider longer sleep time ? For example `self.poll_sleep` because this fix should behave better when openstack is overwhelmed and response from server is 500. Should not we wait a little ?

pvoborni: What wait time would you propose?

Tiboris: To respect timeout to provision all the systems it should be either smaller than that or to use `self.poll_sleep` or longer `self.poll_sleep_initial`. Does it make sense or it does not matter in this particular case?

pvoborni: nova.servers.create just issues a creation of a server. I'm not sure if the timeout makes a big difference. Under load, the issuance might both fail or succeed the next second or it might also fail if we wait 10 mins. 

Maybe it could be reasonable to wait e.g. 10s assuming that there was a short spike in load which can be gone.

Tiboris: One more thing about this (using self.poll_sleep) is that i wanted to get rid of "magic" numbers.

netoarmando: Moved that to a constant as we discussed on IRC.

Tiboris (approved):
Thanks for update, let us merge it.

PR #86, opened 2021-02-10
feat: Add RPM spec file

netoarmando:
Versioning automated via Github Actions.
Dependencies have no pinned version.

### DIFF
--- a/mrack.spec
+++ b/mrack.spec
@@ -1,0 +1,57 @@
+# Created by pyp2rpm-3.3.5
+%global srcname mrack
+
+Name:           %{srcname}
+Version:        0.6.0
+Release:        1%{?dist}
+Summary:        Multicloud use-case based multihost async provisioner for CIs and testing during development
+
+License:        Apache License 2.0
+URL:            https://github.com/neoave/mrack
+Source0:        https://github.com/neoave/mrack/releases/download/v%{version}/mrack-%{version}.tar.gz
+BuildArch:      noarch
+
+BuildRequires:  python3-devel
+BuildRequires:  python3-AsyncOpenStackClient
+BuildRequires:  beaker-client
+BuildRequires:  python3-boto3
+BuildRequires:  python3-botocore
+BuildRequires:  python3-click
+BuildRequires:  python3-pyyaml
+BuildRequires:  python3-setuptools
+
+%description
+mrack is a provisioning library for CI and local multi-host testing supporting multiple
+provisioning providers (e.g. OpenStack, Beaker). But in comparison to other multi-cloud
+libraries, the aim is to be able to describe host from application perspective.
+
+%{?python_provide:%python_provide %{srcname}}
+
+Requires:       python3-AsyncOpenStackClient
+Requires:       beaker-client
+Requires:       python3-boto3
+Requires:       python3-botocore
+Requires:       python3-click
+Requires:       python3-pyyaml
+
+%prep
+%autosetup -n %{srcname}-%{version}
+# Remove bundled egg-info
+rm -rf %{srcname}.egg-info
+
+%build
+%py3_build
+
+%install
+%py3_install
+
+%files -n %{srcname}
+%license LICENSE
+%doc README.md
+%{_bindir}/mrack
+%{python3_sitelib}/%{srcname}
+%{python3_sitelib}/%{srcname}-%{version}-py%{python3_version}.egg-info
+
+%changelog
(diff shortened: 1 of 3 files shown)


0001-chore-set-boto-minimum-versions.patch
```diff
@@ -0,0 +1,32 @@
...
+
+diff --git a/requirements.txt b/requirements.txt
+index ab2705f..9521dbe 100644
+--- a/requirements.txt
++++ b/requirements.txt
+@@ -1,6 +1,6 @@
+ beaker-client>=28
+-botocore==1.18.18
+-boto3==1.15.18
++botocore>=1.14.17
```
pvoborni, line 25: The patch will not be applicable once we release a new version with the amended requirements.txt.

mrack.spec
```diff
@@ -0,0 +1,61 @@
...
+License:        Apache License 2.0
+URL:            https://github.com/neoave/mrack
+Source0:        https://github.com/neoave/mrack/releases/download/v%{version}/mrack-%{version}.tar.gz
+BuildArch:      noarch
+
+### Patches ###
+Patch0001:  0001-chore-set-boto-minimum-versions.patch
+
+BuildRequires:  python3-devel
+BuildRequires:  python3dist(asyncopenstackclient) >= 0.8.1
```
pvoborni, line 18: If I understand https://docs.fedoraproject.org/en-US/packaging-guidelines/Python/ correctly, packager has an option to specify rpm deps directly e.g. python3-asyncopenstackclient or use this - "automatically generated dependencies". When automatically generated deps are used then the requiremnts.txt is checked. This spec file is using version 0.6.0 so it gets it from the released version and not git repo - thus the reason for the Patch0001. 

So I think we have the following options:
- do not use automatically generated deps 
- use the automatically generarated deps, but have corrected requirements.txt (which will happen only after this patch merge and a release).

mrack.spec
```diff
@@ -0,0 +1,61 @@
...
+BuildRequires:  python3dist(beaker-client) >= 28
+BuildRequires:  python3dist(boto3) >= 1.11.17
+BuildRequires:  python3dist(botocore) >= 1.14.17
+BuildRequires:  python3dist(click) >= 7
+BuildRequires:  python3dist(pyyaml) >= 5
+BuildRequires:  python3dist(setuptools)
+
+%description
+mrack is a provisioning library for CI and local multi-host testing
+supporting multiple provisioning providers (e.g. OpenStack, libvirt, containers,
```
pvoborni, line 28: we should probably not mention the stuff which is not-yet implemented (libvirt, containers)

mrack.spec
```diff
@@ -0,0 +1,61 @@
+# Created by pyp2rpm-3.3.5
+%global srcname mrack
+
+Name:           %{srcname}
+Version:        0.6.0
```
pvoborni, line 5: We need a way to have this automatically bumped in a release process.

netoarmando: This can be achieved by adding this to `[tool.semantic_release]` section of the `pyproject.toml` file:
```ini
version_pattern = [
    "mrack.spec:Version:\\s+{version}",
]
```
But I'm still trying to automate the changelog portion.

mrack.spec
netoarmando: "do not use automatically generated deps" -- I preferred this approach.

netoarmando:
This is ready for review, better automation will come after this (e.g. bump changelog, trigger copr).

I've tested a `0.7.0.dev` version with all deps and this on Fedora 33 provisioning hosts with AWS and OpenStack.

pvoborni:
Final review is on Tibor, on high level, I did not spot anything problematic.

mrack.spec
```diff
@@ -0,0 +1,57 @@
...
+Summary:        Multicloud use-case based multihost async provisioner for CIs and testing during development
+
+License:        Apache License 2.0
+URL:            https://github.com/neoave/mrack
+Source0:        https://github.com/neoave/mrack/releases/download/v%{version}/mrack-%{version}.tar.gz
+BuildArch:      noarch
+
+BuildRequires:  python3-devel
+BuildRequires:  python3-AsyncOpenStackClient
+BuildRequires:  beaker-client
```
Tiboris, line 16: Which version of this has the f33?

netoarmando: `beaker-client-28.2-1.fc33.noarch`
All versions defined in `requirements.txt` are still required and checked during package installation.

Tiboris (approved):
Thank you @netoarmando for the patch let us get it merged.

PR #89, opened 2021-02-19
feat: Support size definition in metadata

Tiboris:
feat: Support size definition in metadata

### DIFF
--- a/src/mrack/transformers/transformer.py
+++ b/src/mrack/transformers/transformer.py
@@ -85,9 +85,18 @@ class Transformer:
         return image
 
     def _get_flavor(self, host):
-        """Get flavor by host group."""
-        # TODO: add sizes
-        flavor = get_config_value(self.config["flavors"], host["group"])
+        """
+        Get VM flavor from host metadata definition or based on host group.
+
+        Get the flavor from 'size' definition in host metadata.
+        If 'size' is not specified, get the flavor based on host groups
+        which are defined in provisioning config.
+        """
+        if host.get("size"):  # allow to override size by pointing to group in metadata
+            flavor = get_config_value(self.config["flavors"], host["size"])
+        else:  # default action based on host group define the flavor of instance
+            flavor = get_config_value(self.config["flavors"], host["group"])
+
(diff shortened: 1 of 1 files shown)


pvoborni:
Hi, 

This falls into category of: provider specific configuration in a host configuration. It makes it difficult to simply switch a provider to something else as other providers might not have a concept of flavor. In other words it goes against philosophy of the project: that we have a common configuration in provisioning config and job metadata describes it from application perspective. In the end more difficult to run the same job in different environments.

Better approach would be to identify why we need the change of flavor (the app perspective) and model it in job metadata in a way that it supports switching between provider. 

E.g. we need a bigger machine, that can mean e.g.:
* x-large flavor in openstack
* y-large in AWS 
* 4vcpus and 8Gib in libvirt

Atm it is modeled through group(s). 

Before Mrack, the limitation was that only one group per host was possible. Now mrack can have multiple groups. The only issue is that we don't know which group is the one to pick for flavor. 

But instead a 'flavor', I can imaging that we introduce a new host attribute called 'size'. It reflects the app usage. 

So e.g. we can have a provisioning config which has:
```yaml
openstack:
  flavors:
     ipaserver: m1.medium
     ipaclient: m1.micro
     large: m1.large
     huge: m1.xlarge
aws:
  flavors:
     ipaserver: t1.medium
     ipaclient: t1.micro
     large: t1.large
     huge: t1.xlarge
```
Then job metadata file can have:

```yaml
domains: 
   name: foo.bar
   hosts: 
      - name: hugeserver.foo.bar
        group: ipaserver
        size: huge
     - name: normalserver.foo.bar
       group: ipaserver
```

This will work even with provider switching. 


That said, I'm not against having a support for per-provider override in Mrack. But I'd rather if it was more explicit and really per-provider and that the it would have general build-in support. My initial ideas, which were not yet implemented were that we will have 3 sources of configuration for a host - the closer the higher priority:

- a provisioning config
- job metadata file
- host definition

I can dump it into a design page if needed.

netoarmando (approved):

PR #90, opened 2021-03-12
chore: change boto's minimum version required

netoarmando:
Replace exact versions of boto3 and boto-core with "greater than" and setting an older version because it's the latest available for Fedora 32.

It will install latest via pypi, pinned version is not a requirement anymore because production conflicts with linchpin are not a problem anymore.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,6 +1,6 @@
 beaker-client>=28.1
-botocore==1.18.18
-boto3==1.15.18
+botocore>=1.14.17
+boto3>=1.11.17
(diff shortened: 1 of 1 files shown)


Tiboris:
This minimal requirement IMO should not cause any breakage. It will only affect systems that may not install latest and greatest boto. Thus ACK.

Tiboris (approved):
This minimal requirement IMO should not cause any breakage. It will only affect systems that may not install latest and greatest boto. Thus ACK.

PR #94, opened 2021-03-26
chore: Trigger COPR to build rpm

netoarmando:
Add action step to trigger COPR after a new release is created.

Using Github's built-in webhooks failed to talk to COPR "after release", so this should be more reliable.

--- 
**TO DO:**
- [x] Add secret to repo

### DIFF
--- a/.github/workflows/test-and-release.yml
+++ b/.github/workflows/test-and-release.yml
@@ -66,3 +66,5 @@ jobs:
         with:
           github_token: ${{ secrets.GITHUB_TOKEN }}
           pypi_token: ${{ secrets.PYPI_TOKEN }}
+      - name: Trigger COPR build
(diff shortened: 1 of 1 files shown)


Tiboris (approved):
I have no problem with this way of triggering.
Just a note, if the semantic release will not release newer version (commits are only chore, refactor, etc.) the pipeline result will be green and release will not be created.
See:https://github.com/neoave/mrack/actions/runs/679553191

netoarmando:
> I have no problem with this way of triggering.
> Just a note, if the semantic release will not release newer version (commits are only chore, refactor, etc.) the pipeline result will be green and release will not be created.
> See:https://github.com/neoave/mrack/actions/runs/679553191

That's a good point, but we are ok, for those cases a new build will be triggered, but based on the same release.

Merging this, we can revisit it in the future.

PR #95, opened 2021-03-26
docs: Add info on how to install on Fedora

netoarmando:
Packages availble as rpm for Fedora 32+


Tiboris (approved):
Thanks for the update @netoarmando

PR #96, opened 2021-03-30
feat: Advanced podman support and several minor updates and bugfixes

Tiboris:
## feat: Add capability to use custom podman options

mrack now support advanced podman provisioning
to enable custom way of running pods with podman
we should add podman section to provisioning config
something like:

```
podman:
    images:
        # systemd enabled container images from: https://github.com/Tiboris/snappeas
        fedora-32: tdudlak/snappeas:fedora-32
        fedora-33: tdudlak/snappeas:fedora-33
        fedora-rawhide: tdudlak/snappeas:fedora-rawhide

    pubkey: config/id_rsa.pub

    default_network: mrack-network

    podman_options:
        # Mount a temporary filesystems (tmpfs) into a container
        "--tmpfs":
            - "/tmp"
        # Use /sys/fs/cgroup in container as read only volume
        "-v":
            - "/sys/fs/cgroup:/sys/fs/cgroup:ro"
        # Adding ipv6 support to network
        "--network": "enable_ipv6=true"
```

## feat: Use ssh public key to access container instead of podman id

Use ssh to connect to container provisioned by mrack
instead of ansible connection to prevent some issues
with connection and built in ansible modules so
the container acts like a vm.

## feat: Use more flexible way of defining podman names
    
NETWORK: Podman now use network name created by composing
default network from provisioning config and domain name.
PODS: name of the pod derives from the hostname and network
the pod is assigned to.

### DIFF
--- a/src/mrack/providers/podman.py
+++ b/src/mrack/providers/podman.py
@@ -59,7 +59,7 @@ class PodmanProvider(Provider):
     async def create_server(self, req):
         """Request and create resource on selected provider."""
         hostname = req["name"]
-        logger.info(f"Creating container for host: {hostname}")
+        logger.info(f"{self.dsp_name}: Creating container for host: {hostname}")
         hostname = req["name"]
         image = req["image"]
         network = req.get("network")
@@ -78,7 +78,7 @@ class PodmanProvider(Provider):
             try:
                 inspect = await self.podman.inspect(cont_id)
             except ProvisioningError as err:
-                logger.error(object2json(err))
+                logger.error(f"{self.dsp_name}: {object2json(err)}")
                 raise ServerNotFoundError(cont_id) from err
             server = inspect[0]
             running = server["State"]["Running"]
@@ -92,12 +92,15 @@ class PodmanProvider(Provider):
 
         if datetime.now() >= timeout_time:
             logger.warning(
-                f"{cont_id} was not provisioned within a timeout of" f" {timeout} mins"
+                f"{self.dsp_name}: {cont_id} was not provisioned within a timeout of"
+                f" {timeout} mins"
             )
         else:
-            logger.info(f"{cont_id} was provisioned in {prov_duration:.1f}s")
+            logger.info(
+                f"{self.dsp_name}: {cont_id} was provisioned in {prov_duration:.1f}s"
+            )
 
-        logger.debug(object2json(server))
+        logger.debug(f"{self.dsp_name}: Resource: {object2json(server)}")
         return server
 
     async def delete_host(self, host_id):
--- a/src/mrack/providers/podman.py
+++ b/src/mrack/providers/podman.py
@@ -16,6 +16,7 @@
 
 import asyncio
 import logging
+import os
 from datetime import datetime, timedelta
 
 from mrack.errors import ProvisioningError, ServerNotFoundError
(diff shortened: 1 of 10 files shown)


src/mrack/providers/utils/podman.py
```diff
@@ -81,6 +102,38 @@ async def rm(self, container_id, force=False):  # pylint: disable=invalid-name
...
+        args = ["stop"]
+        if time:
+            args.append("--time")
+            args.append(time)
+
+        args.append(container_id)
+        _stdout, _stderr, process = await self._run_podman(args, raise_on_err=False)
+        return process.returncode == 0
+
+    async def network(self, network, exists=True):
```
netoarmando, line 123: This method behaves in a strange way, calling it with a non-existing network with `exists=False`, to ensure it doesn't exist (deleting if it exists), only calls `podman inspect {network}` and that's not clear.

Why not split this into `create_network` and `delete_network` and make them idempotent methods?
Another option: change argument to `action='create'` and actions might be `create` or `remove` as well.

netoarmando (changes requested):
`src/mrack/data/mrack_seccomp.json` was included here, must it be copied to "work dir" before calling `mrack up`?

src/mrack/transformers/podman.py
```diff
@@ -35,6 +36,7 @@ async def init_provider(self):
 
         await self._provider.init(
             container_images=self.config["images"].values(),
+            ssh_key=self.config["pubkey"],
```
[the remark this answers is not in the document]
Tiboris: Yup correct we can update beaker in one more commit.

src/mrack/providers/utils/podman.py
Tiboris: yes i can try something and push

src/mrack/transformers/podman.py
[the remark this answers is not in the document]
Tiboris: Done

netoarmando:
@Tiboris Testing I found that mrack ssh is not working, this is needed: 
```diff
diff --git a/src/mrack/actions/ssh.py b/src/mrack/actions/ssh.py
index f6b6660..4bee9d4 100644
--- a/src/mrack/actions/ssh.py
+++ b/src/mrack/actions/ssh.py
@@ -127,7 +127,7 @@ class SSH:
         """Simulate SSH by attaching an interactive session to a container."""
         if host.provider.name == "podman":
             podman = Podman()
-            podman.interactive(host.id)
+            podman.interactive(host.host_id)
         else:
             raise NotImplementedError("Docker is not yet supported.")
```

After it as non-root I could ssh into the container. However, to use 'plain' ssh I had to call mrack as root (and comment a few lines), as already discussed. 

Also based on current container image (fedora-rawhide), its user is different (root) from the one in provisioning-config (fedora) for that image name. I guess we need to suffix names with `-container` or something else. 

Rest looks good to me.

Tiboris:
> @Tiboris Testing I found that mrack ssh is not working, this is needed:
> ...

Nice catch @netoarmando !

src/mrack/providers/podman.py
```diff
@@ -92,17 +122,46 @@ async def wait_till_provisioned(self, resource):
...
 
-        logger.debug(object2json(server))
         return server
 
     async def delete_host(self, host_id):
         """Delete provisioned host."""
-        deleted = await self.podman.rm(host_id, force=True)
+        # first we inspect the container to find its networks
+        insp_data = await self.podman.inspect(host_id)
+        networks = insp_data[0]["NetworkSettings"]["Networks"]
```
netoarmando, line 157: Are all networks take into account here? That means this deletes all networks, let's say one is using mrack and has a previous network in their machine, will this destroy that network as well?

netoarmando: Nice, got it.

netoarmando (approved):

PR #97, opened 2021-04-15
feat: use ssh to check VM availability after provisioning

Tiboris:
After a host in provisioned we do check ssh connectivity
to the host itself and claim it is ready only after ssh
connection is successful. If ssh connection fails we
mark this host as an error host delete it and try to
go with reprovision strategy.

### DIFF
--- a/src/mrack/utils.py
+++ b/src/mrack/utils.py
@@ -21,13 +21,18 @@ import datetime
 import json
 import logging
 import os
+import subprocess
 import sys
+from typing import Dict
 
 import yaml
 
 from mrack.errors import ConfigError
 
 logger = logging.getLogger(__name__)
+
+# Singleton context holder
+global_context: Dict = {}
 
 
 def get_config_value(config_dict, key, default=None):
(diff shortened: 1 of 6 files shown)


src/mrack/utils.py
```diff
@@ -201,6 +206,51 @@ def get_ssh_key(host, meta_host, config):
...
+        cmd.extend(["-i", ssh_key])
+    psw_input = None
+    if password and not ssh_key:
+        cmd.extend(["-o", "'PasswordAuthentication'"])
+        psw_input = f"{password}\n"
+
+    cmd.append(host.ip_addr)  # Destination
+
+    if execute:
+        cmd.extend(["-C", execute])
```
netoarmando, line 244:
```
$ ssh [-options] destination [command]
```
`-C` "Requests compression of all data... Compression is desirable on modem lines and other slow connections, but will only slow down things on fast networks."
Should we remove this?

src/mrack/utils.py
```diff
@@ -201,6 +206,51 @@ def get_ssh_key(host, meta_host, config):
     return ssh_key
 
 
+def ssh_to_host(
+    host,
+    execute=None,
```
netoarmando, line 211: IMHO `command` is a better name, also I think we should pass this as last argument.

netoarmando (approved):
LGTM, tested with 1 and 10 hosts, I haven't hit any SSH error, but logs showed that this is working properly. I left comments for two nit-picks.

Issue #98, opened 2021-04-16
Mrack does not log error when ssh fails after active state for machine


Tiboris:
mrack.log:
```
2021-04-16T15:12:28     Warning: Permanently added '10.0.104.154' (ECDSA) to the list of known hosts.

2021-04-16T15:12:28     cloud-user@10.0.104.154: Permission denied (publickey,gssapi-keyex,gssapi-with-mic).

2021-04-16T15:12:28     ssh -o 'StrictHostKeyChecking=no' -o 'UserKnownHostsFile=/dev/null' -l cloud-user -i config/id_rsa 10.0.104.132 -C echo mrack

2021-04-16T15:12:28     Warning: Permanently added '10.0.104.132' (ECDSA) to the list of known hosts.

2021-04-16T15:12:28     cloud-user@10.0.104.132: Permission denied (publickey,gssapi-keyex,gssapi-with-mic).

2021-04-16T15:12:28     ssh -o 'StrictHostKeyChecking=no' -o 'UserKnownHostsFile=/dev/null' -l cloud-user -i config/id_rsa 10.0.107.21 -C echo mrack

2021-04-16T15:12:28     Warning: Permanently added '10.0.107.21' (ECDSA) to the list of known hosts.

2021-04-16T15:12:28     OpenStack: 2 hosts were not provisioned properly, deleting.

2021-04-16T15:12:28     OpenStack: Error: None

2021-04-16T15:12:28     OpenStack: Error: None
```

PR #99, opened 2021-04-19
fix: missing error logging while checking the ssh connection

Tiboris:
Aims to resolve https://github.com/neoave/mrack/issues/98

After this patch:
```
2021-04-19 12:56:24,432 mrack.utils INFO ssh -o 'StrictHostKeyChecking=no' -o 'UserKnownHostsFile=/dev/null' -l fedora -i config/id_rsa 10.0.99.34 echo mrack
2021-04-19 12:56:24,456 mrack.providers.provider INFO OpenStack: 1 hosts were not provisioned properly, deleting.
2021-04-19 12:56:24,457 mrack.providers.provider ERROR OpenStack: Error: Could not establish ssh connection to host d0160fba-803e-4ed8-bab4-4b42c248849d with IP 10.0.99.34
```

### DIFF
--- a/src/mrack/providers/provider.py
+++ b/src/mrack/providers/provider.py
@@ -76,21 +76,29 @@ class Provider:
             TimeoutError: The port isn't accepting connection after specified `timeout`.
         """
         start_time = datetime.now()
+        info_msg = (
+            f"{self.dsp_name}: Waiting for the port {port} "
+            f"on host {host.ip_addr} to start accepting connections"
+        )
+        logger.info(info_msg)
+
         while True:
             try:
                 with socket.create_connection((host.ip_addr, port), timeout=timeout):
+                    logger.info(
+                        f"{self.dsp_name}: Port {port} on host "
+                        f" {host.ip_addr} is now open"
+                    )
                     break
             except OSError:
                 await asyncio.sleep(10)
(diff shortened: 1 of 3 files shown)


src/mrack/providers/provider.py
```diff
@@ -224,9 +232,17 @@ async def parse_error_hosts(self, hosts):
...
             )
 
+            if host.status != STATUS_ACTIVE:
+                errors.append(host)
+                continue
+
             # self._wait_for_ssh checks the host's capability to be
             # connected to using ssh; if the check fails we reprovision
-            if host.status != STATUS_ACTIVE or not await self._wait_for_ssh(host):
+            if not await self._wait_for_ssh(host):
```
netoarmando, line 241: If host is not active we should not check its connectivity, right?

Tiboris: I thought this continue helps https://github.com/neoave/mrack/pull/99/files#diff-9eb08e19e12ca58d0f3b3a61d2426fbd72ebefbd5e7171981e5fa4cbac237e78R237

netoarmando: You are right, I overlooked the for loop.

netoarmando (approved):

PR #101, opened 2021-04-22
fix: ssh action crashes as it misses global_context

pvoborni:
utils.ssh_to_host requires to have global_context initialized but
with introduction of it ssh action was not ammended. Without it
mrack crashed on sshing to a host.

This fix is initializing utils.global_context in ssh_action and
thus fixing the issue.


Tiboris (approved):
Thanks for the fix @pvoborni

PR #103, opened 2021-04-22
feat: Pull podman images and create networks in preparation

Tiboris:
Pull podman image in provisioning preparation
to not block the provisioning or pull one image
more than once which may cause unexpected errors.

Create podman networks in provisioning preparation
to not block the provisioning or ty to create already
created networks which may cause unexpected errors.

### DIFF
--- a/src/mrack/providers/podman.py
+++ b/src/mrack/providers/podman.py
@@ -58,11 +58,36 @@ class PodmanProvider(Provider):
         logger.info(f"{self.dsp_name}: Init duration {login_duration}")
 
     async def prepare_provisioning(self, reqs):
-        """Pull missing images to prepare provisioning for podman."""
+        """Prepare podman network and pull missing images to prepare provisioning."""
         image_list = set()
+        network_list = set()
 
         for req in reqs:
+            # First create list of podman network(s) to be created
+            # podman network(s) should be always created no matter what
+            network = req.get(
+                "network", f"{self.default_network}-{req['domain'].replace('.', '-')}"
+            )
+            # store the used network per requirement to be later accessed
+            req["network"] = network
+            network_list.add(network)
+
+            # Prepare image_list to be pulled later
             image_list.add(req["image"])
+
+        if network_list:
+            logger.info(f"{self.dsp_name}: Preparing network(s) {network_list}")
+            awaitables = []
+            for network in network_list:
+                if not await self.podman.network_exists(network):
+                    awaitables.append(self.podman.network_create(network))
+
+            network_results = await asyncio.gather(*awaitables)
+            success = all(network_results)
+
+            if not success:
+                logger.error(f"{self.dsp_name}: Creation of networks failed")
+                return False
 
         logger.info(f"{self.dsp_name}: Pulling missing images {image_list}")
 
@@ -96,10 +121,12 @@ class PodmanProvider(Provider):
         logger.info(f"{self.dsp_name}: Creating container for host: {hostname}")
 
         image = req["image"]
-        network = req.get(
-            "network", f"{self.default_network}-{req['domain'].replace('.', '-')}"
-        )
-        await self.podman.network_create(network)
+        network = req.get("network")  # preparation method should set this value
+        if not network:
+            logger.error(
+                f"{self.dsp_name}: Failed to load network requirement from: {req}"
+            )
+            raise ProvisioningError("Could not set up podman network for some host(s)")
 
         container_id = await self.podman.run(
             image,
--- a/src/mrack/providers/podman.py
+++ b/src/mrack/providers/podman.py
@@ -58,8 +58,29 @@ class PodmanProvider(Provider):
         logger.info(f"{self.dsp_name}: Init duration {login_duration}")
 
     async def prepare_provisioning(self, reqs):
-        """Prepare provisioning."""
-        pass
+        """Pull missing images to prepare provisioning for podman."""
+        image_list = set()
+
+        for req in reqs:
+            image_list.add(req["image"])
+
+        logger.info(f"{self.dsp_name}: Pulling missing images {image_list}")
+
+        awaitables = []
+        for image in image_list:
+            if not await self.podman.image_exists(image):
+                logger.debug(f"{self.dsp_name}: Pull of image '{image}' required")
+                awaitables.append(self.podman.pull(image))
+
+        pull_results = await asyncio.gather(*awaitables)
+        success = all(pull_results)
+
+        if success:
+            logger.info(f"{self.dsp_name}: All required images present")
+        else:
+            logger.error(f"{self.dsp_name}: Pulling of missing images failed")
+
+        return success
 
     async def validate_hosts(self, reqs):
         """Validate that host requirements are well specified."""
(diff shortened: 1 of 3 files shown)


src/mrack/providers/podman.py
```diff
@@ -58,8 +58,33 @@ async def init(self, container_images, default_network, ssh_key, container_optio
...
+        image_list.add("quay.io/neoave/snappeas:fedora-33")
+        for req in reqs:
+            image = req["image"]
+            if not await self.podman.image_exists(image):
+                logger.debug(f"{self.dsp_name}: Pull of image '{image}' required")
+                image_list.add(image)
+            else:
+                logger.debug(f"{self.dsp_name}: Image '{image}' present")
+
+        if not image_list:
```
pvoborni, line 72: The success message that we have all the images could be moved after the `success = all(pull_results)`.  So that there would be some pointer that this part of operation ended.

src/mrack/providers/podman.py
```diff
@@ -58,8 +58,33 @@ async def init(self, container_images, default_network, ssh_key, container_optio
         logger.info(f"{self.dsp_name}: Init duration {login_duration}")
 
     async def prepare_provisioning(self, reqs):
-        """Prepare provisioning."""
-        pass
+        """Pull missing images to prepare provisioning for podman."""
+        image_list = set()
+        image_list.add("quay.io/neoave/snappeas:fedora-33")
```
pvoborni, line 63: Is having ` image_list.add("quay.io/neoave/snappeas:fedora-33")` here intentional? Looks as leftover from testing.

pvoborni (commented):
Thanks for the PR. Some comments added. 

A minor thing for next time: it would also good if the unrelated changes of logging messages would be in a separate commit.

src/mrack/providers/podman.py
Tiboris: Correct

src/mrack/providers/podman.py
Tiboris: Can do

src/mrack/providers/podman.py
```diff
@@ -58,17 +58,42 @@ async def init(self, container_images, default_network, ssh_key, container_optio
...
 
         for req in reqs:
+            # First create list of podman network(s) to be created
+            # podman network(s) should be always created no matter what
+            network = req.get(
+                "network", f"{self.default_network}-{req['domain'].replace('.', '-')}"
+            )
+            # store the used network per requirement to be later accessed
+            req["network"] = network
+            if not await self.podman.network_exists(network):
```
pvoborni, line 73: `self.podman.network_exists(network)` check is run for all reqs no matter what in sequence. So e.g. for 5 machines with the same network, it is run 4 times unnecessarily. 

Better logic would be:
- generate the list of network names in `network_list` (and as it is a set, it will have only unique)
- run `self.podman.network_exists(network):` for each network in network_list
- create the ones which doesn't exist

Tiboris: Correct, will update.

Tiboris: Done

src/mrack/providers/podman.py
```diff
@@ -58,8 +58,58 @@ async def init(self, container_images, default_network, ssh_key, container_optio
...
+            network = req.get(
+                "network", f"{self.default_network}-{req['domain'].replace('.', '-')}"
+            )
+            # store the used network per requirement to be later accessed
+            req["network"] = network
+            network_list.add(network)
+
+            # Prepare image_list to be pulled later
+            image = req["image"]
+            if not await self.podman.image_exists(image):
```
pvoborni, line 77: This 'if' should be on line 102 as I wrote earlier.  Where it woudl check it only once per image. 

So that line 100 would be like:
```python
        awaitables = []
        for image in image_list:
            if not await self.podman.image_exists(image):
                 awaitables.append(self.podman.pull(image))
```

Tiboris: Updated

bhavikbhavsar:
Following scenarios where tested:

1.  Defined multiple hosts in the metadata for different OS, pre-setup pull only one image.
https://pastebin.com/UhAqh9rZ - success
Only missing image is pulled.
Hosts within the same domain are able to communicate.
Qs:
 - what if user wants to have the latest image pulled whenever mrack provisions?

2. When url for image is incorrect in provisioning-config
https://pastebin.com/j0hZ21fv - success
Proper error is displayed

3. When multiple domains defined in metadata.
https://pastebin.com/9aG8DrKz - Fails
Fails after creating the first two host from first domain as hostname are treated as same.
Qs:
- What should be default behaviour here, is it necessary for user to define separate hostname?

4. Define different hostname for each domain.
https://pastebin.com/YkdZATHv 
Current behaviours different network ips are assigned as per the domains. But all provisioned containers are able to communicate with one another.
Qs:
- What is the expected behaviour, when provisioned in two different domains? Should it be able to communicate between different domains or it should not?

pvoborni:
> Following scenarios were tested:
> 
>     1. Defined multiple hosts in the metadata for different OS, pre-setup pull only one image.
>        https://pastebin.com/UhAqh9rZ - success
>        Only missing image is pulled.
>        Hosts within the same domain are able to communicate.
>        Qs:
> 
> 
>     * what if user wants to have the latest image pulled whenever mrack provisions?

Don't know if we need it. But can be an RFE if needed.

> 
> 
>     1. When url for image is incorrect in provisioning-config
>        https://pastebin.com/j0hZ21fv - success
>        Proper error is displayed
> 
>     2. When multiple domains defined in metadata.
>        https://pastebin.com/9aG8DrKz - Fails
>        Fails after creating the first two host from first domain as hostname are treated as same.
>        Qs:
> 
> 
>     * What should be default behaviour here, is it necessary for user to define separate hostname?

It is expected in Mrack that each name of a host will be unique, regardless of domain. In such case it is better to use FQDNs. I don't know what is the error in such case, but it might be good RFE to fail early in such situation. In any case it is IMO not related to this PR. 

> 
> 
>     1. Define different hostname for each domain.
>        https://pastebin.com/YkdZATHv
>        Current behaviours different network ips are assigned as per the domains. But all provisioned containers are able to communicate with one another.
>        Qs:
> 
> 
>     * What is the expected behaviour, when provisioned in two different domains? Should it be able to communicate between different domains or it should not?

They should be able to communicate. E.g. for testing cross realm trusts (not supported now in IPA, but an example).

> 
> 
> Also a general observation between each mrack up had to delete mrackdb.json. It becomes annoying. Can a hashmap be created to record each run in db? So each time db file need not be deleted.

I'd rephrase that if in the **same** working directory, host definition in **metadata changes**, especially in names of hosts, then it is required to delete the DB file. It is not required if metadata doesn't change.  I guess we can define, what would be a good behavior in such case. In either supporting it better or at least writing better errors. That said, I'd recommend to have a separate working dir for different test cases.

Tiboris:
> > Following scenarios were tested:
> > ```
> > 1. Defined multiple hosts in the metadata for different OS, pre-setup pull only one image.
> >    https://pastebin.com/UhAqh9rZ - success
> >    Only missing image is pulled.
> >    Hosts within the same domain are able to communicate.
> >    Qs:
> > 
> > 
> > * what if user wants to have the latest image pulled whenever mrack provisions?
> > ```
> 
> Don't know if we need it. But can be an RFE if needed.

I like this idea, probably some option in provisioning-config.yaml to re pull if image exists.

pvoborni (approved):
Code looks good to me. Approving - referring to Bhavik's testing.

PR #105, opened 2021-04-23
fix: Do not add the config_drive_req to req

Tiboris:
Do not add the config_drive_req to req when
the requirement is not needed to behave as older
mrack releases.


bhavikbhavsar:
Currently we don't see this issue lets reopen this pull request if the issue crops up again.

PR #106, opened 2021-04-27
feat: Use GlobalContext class

Tiboris:
Use GlobalContext class in runtime
which allows us to use mrack with only this context
and as a library with only creating desired actions
Up/Destroy/... and calling them in python scripts.

some minor fixes are also included in this PR

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,8 +2,6 @@
 
 ![pypi_badge](https://img.shields.io/pypi/v/mrack?label=PyPI&logo=pypi) ![readthedocs_badge](https://img.shields.io/readthedocs/mrack?label=Read%20the%20Docs&logo=read-the-docs) ![badge](https://copr.fedorainfracloud.org/coprs/g/freeipa/neoave/package/mrack/status_image/last_build.png)
 
-**Important**: most of the described below is not implemented yet
-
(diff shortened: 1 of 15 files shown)


pvoborni:
Before I'll read the code: Why is this "breaking" change needed?

Tiboris:
@pvoborni Hello, please do not. I need to update this on more places. and add at least one commit.
> Why is this "breaking" change needed?

I was trying how semantic-release will behave - after merge next release would be 1.0.0

pvoborni:
> Why is this "breaking" change needed?

The emphasis is not on the "breaking". The question is why is this PR needed? What benefits we expect from it?

Tiboris:
Tested:
```
mrack up -m metadata-f30.yaml
mrack ssh -m metadata-f30.yaml
mrack destroy -m metadata-f30.yaml
```

netoarmando (commented):
I agree with implementation, this would make communication between components simpler without compromising separation of concerns.  Bonus for integrating the singleton with clicky context.

However, it would be nice to include some comments on behavior or even some docs on this.

Tiboris:
Will do @netoarmando

netoarmando (approved):
LGTM! Thanks for updating the documentation.

PR #108, opened 2021-04-30
fix: The type of duration is not correct

Tiboris:
TypeError: unsupported format string passed to datetime.timedelta.__format__


bhavikbhavsar (approved):
Looks good to merge

PR #109, opened 2021-04-30
feat: Add --version option to mrack cli

Tiboris:

### DIFF
--- a/src/mrack/run.py
+++ b/src/mrack/run.py
@@ -14,6 +14,8 @@
 
 """mrack default app."""
 #  pylint: disable=no-name-in-module
+__version__ = "0.10.0"
+
 import asyncio
 import logging
 import os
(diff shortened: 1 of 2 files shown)


src/mrack/run.py
```diff
@@ -14,6 +14,8 @@
 
 """mrack default app."""
 #  pylint: disable=no-name-in-module
+__version__ = "0.10.0"
```
pvoborni, line 17: shouldn't it be removed also from `__init__.py`? In general +1 for moving it. But wouldn't it be better to be moved to a special file for it. E.g. version.py. I don't think it belongs into run.py. In the future, when we somehow stabilize API, Mrack could be used as a library and run.py would not be relevant in such use case.

pvoborni (changes requested):

src/mrack/run.py
Tiboris: Given the nature of scripts/mrack.py where importing would create circular import I have decided that I will add another __version__ variable where we need it, to the run.py.
We have well automated the version update so I hope having it on 2 places, for CLI and mrack itself is not a big issue.
I was used to the fact the __version__ variable is in __init__.py

Tiboris:
Closing in favor of https://github.com/neoave/mrack/pull/112

PR #110, opened 2021-04-30
chore: accept code which has 10 lines or less similar

Tiboris:


pvoborni (changes requested):
Code looks and works good, but the commit title is misleading. Should be the opposite e.g., "chore: accept code which has 10 lines or less similar" as currently it is more aggressive.

Tiboris:
Can do

pvoborni (approved):

PR #112, opened 2021-04-30
feat: Add --version option to mrack cli - alternative implementation

pvoborni:
From my PoV, this is cleaner as there is version only on one location in the mrack package and it can be even more easily imported anywhere else. 

The second commit is also logging the version into mrack.log so that it is usable in debugging.

Alternative to: https://github.com/neoave/mrack/pull/109


Tiboris (approved):
Then let us do it this way.

PR #115, opened 2021-05-07
feat: Add capability to disable ssh check in provisioning-config.yaml and update prio of hardcoded Administrator user

Tiboris:
These changes allows us to use older behavior with only changin provisioning-config.yaml file.

### DIFF
--- a/src/mrack/providers/provider.py
+++ b/src/mrack/providers/provider.py
@@ -20,7 +20,7 @@ from datetime import datetime, timedelta
 
 from mrack.errors import ProvisioningError, ValidationError
 from mrack.host import STATUS_ACTIVE, STATUS_OTHER, Host
-from mrack.utils import ssh_to_host
+from mrack.utils import global_context, ssh_to_host
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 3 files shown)


src/mrack/providers/provider.py
```diff
@@ -174,26 +174,32 @@ async def _provision_base(self, reqs):  # pylint: disable=too-many-locals
...
-            if res[RET_CODE]:
-                success_hosts.append(res[HOST_OBJ])
-            else:
-                res[HOST_OBJ].error = (
-                    "Could not establish ssh connection to host "
-                    f"{res[HOST_OBJ].host_id} with IP {res[HOST_OBJ].ip_addr}"
-                )
-                error_hosts.append(res[HOST_OBJ])
+
+        if global_context["config"].get("post_provisioning_ssh_check", True):
```
pvoborni, line 179: I'm not really fan of this. IMHO global config for SSH check should be used in a very rare circumstances.  Providing this only as a global option will lead to disabling it essentially for everything even if it was used only for a certain use case - specific image.  Making ssh_check feature useless.  

Better would be to have it configured per-host with possibility to define on host level, metadata level and provisioning config level. In provisioning config, it might be also a possibility to do it per-image (os).  That way we might have a possibility to disable it only in cases where it is needed. 

I.e. I'd move the logic to transformer.

pvoborni: When thinking about it, introducing global context in providers is also a path to hell. It removes separation of concerns, in long run making it a spaghetti code where everything depends on everything.

pvoborni (changes requested):
The Windows username commit looks OK, but I suggest to use all the usecases we have. We used to break it in the past.

The config drive commit looks OK, though in future we might also have bit more fine-grain configuration (host > metadata > provisioning config)

src/mrack/providers/provider.py
Tiboris: I understand your concern about depending on global context.
I will add a parameter to the method call and fallback to global config if parameter is not set.

pvoborni: I don't understand much the steps you are going to do. So will briefly re-cap: this logic should be in a tranformer as it has access to cofiguration which is provided to it.

f-trivino (approved):
LGTM!

PR #116, opened 2021-05-10
feat(OpenStack): distribute choosing of networks

pvoborni:
Currenlty Mrack always chooses a network with the most available IPs.

This is good for a lot of use cases but is problematic when a lot of jobs are started at the same time. Each of the jobs is doing a
separate network availability check. This can lead to a situation where almost all jobs pick the same and start provisioning with it.

But this network might not have enough IPs for everything so there is a increased chance that it will fail. E.g if there is 300 available IPs and we run 100 jobs which needs 3-4 machines.

Indroducing a new behavior where Mrack is picking a random network from networks which have at least 50% of IPs as the bests one. Thus spreading the load and lowering the chance of the issue to happening.

It will not solve the issue completely. This can be improved e.g. in re-provision logic.

### DIFF
--- a/src/mrack/transformers/openstack.py
+++ b/src/mrack/transformers/openstack.py
@@ -14,6 +14,7 @@
 
 """OpenStack transformer module."""
 import logging
+import random
 
 from mrack.errors import ProvisioningConfigError
 from mrack.transformers.transformer import Transformer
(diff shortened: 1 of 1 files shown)


netoarmando (approved):
Code looks good to me.

PR #119, opened 2021-05-12
feat: Check and wait for resources when provisiong

Tiboris:
feat: Wait for provider resources up to hour

Checking resources more than once should help
to provide better results when we are facing
the providers to be overocupied.

feat(OpenStack): Check openstack response for resources issue

OpenStack may be oversubscribed so even when we do not reach
our quota limit provisioning may fail due to lack of resources.
In such scenario (OpenStack run out of hosts to provision)
we check the response from server and act based on the fact
that OpenStack itself is being fully loaded and does not have
free resources to provide for our server.

docs: Update method docsting as it does not Validate hosts anymore

feat: gracefully destroy servers after traceback

This allows us to create dummy Host object based
on the ProvisioningError traceback arguments
where we added also requirement which may have
caused the provisioning to fail so the worflow
can continue and remove already provisioned
servers after reprovision / abort strategy is
done so resources that may not be freed
to openstack will be returned.

### DIFF
--- a/src/mrack/providers/openstack.py
+++ b/src/mrack/providers/openstack.py
@@ -553,7 +553,8 @@ class OpenStackProvider(Provider):
             if error_attempts > SERVER_ERROR_RETRY:
                 # now we are past to what we would like to wait fail now
                 raise ProvisioningError(
-                    f"{self.dsp_name}: Failed to create server {object2json(specs)}"
+                    f"{self.dsp_name}: Failed to create server {req['name']}",
+                    req,  # add the requirement dictionary to traceback for later
                 )
 
             fault = response["server"].get("fault", {})
--- a/src/mrack/providers/openstack.py
+++ b/src/mrack/providers/openstack.py
@@ -44,8 +44,9 @@ logger = logging.getLogger(__name__)
 
 
 PROVISIONER_KEY = "openstack"
-SERVER_ERROR_RETRY = 5
-SERVER_ERROR_SLEEP = 10
+SERVER_ERROR_RETRY = 5  # number of times to retry server creation
+SERVER_ERROR_SLEEP = 10  # seconds
+SERVER_RES_SLEEP = 10  # minutes
 
 
 class OpenStackProvider(Provider):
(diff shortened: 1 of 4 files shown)


src/mrack/providers/openstack.py
```diff
@@ -545,13 +545,39 @@ async def create_server(self, req):
...
+
+            if error_attempts > SERVER_ERROR_RETRY:
+                # now we are past to what we would like to wait fail now
+                raise ProvisioningError(
+                    f"{self.dsp_name}: Failed to create server {object2json(specs)}"
+                )
+
+            if (
+                response["code"] == 500
+                and "No valid host was found." in response["message"]
```
netoarmando, line 560: Check for 500 is enough, no need to check for string.

Tiboris:
I wonder if we need to increase provisioning timeout from 1hr to at least 2 as with this patch it will be 1hr wait for resources at validation and 1hr wait for server response

src/mrack/providers/provider.py
```diff
@@ -127,7 +127,9 @@ async def _wait_for_ssh(self, host, port=22, timeout=1200):
 
         return res, host
 
-    async def _provision_base(self, reqs):  # pylint: disable=too-many-locals
+    async def _provision_base(
+        self, reqs, res_check_timeout=60
```
netoarmando, line 131: Could you please state in the docstring that `res_check_timeout` default is 60 minutes?

src/mrack/providers/openstack.py
```diff
@@ -545,13 +546,38 @@ async def create_server(self, req):
...
+                    await asyncio.sleep(SERVER_ERROR_SLEEP)
+                    continue  # Try again due to ServerError
+
+            if error_attempts > SERVER_ERROR_RETRY:
+                # now we are past to what we would like to wait fail now
+                raise ProvisioningError(
+                    f"{self.dsp_name}: Failed to create server {object2json(specs)}"
+                )
+
+            fault = response["server"].get("fault")
```
netoarmando, line 559:
```python
fault = response["server"].get("fault", {})
if fault.get("code") == 500:
```

src/mrack/providers/provider.py
Tiboris: Can do

src/mrack/providers/openstack.py
Tiboris: fixed

netoarmando (approved):

netoarmando (commented):
Wait, do not merge it yet..

src/mrack/providers/openstack.py
```diff
@@ -545,13 +546,38 @@ async def create_server(self, req):
...
-                        f"{self.dsp_name}: Fail to create server", specs
-                    ) from exc
-                await asyncio.sleep(SERVER_ERROR_SLEEP)
+                if error_attempts <= SERVER_ERROR_RETRY:
+                    await asyncio.sleep(SERVER_ERROR_SLEEP)
+                    continue  # Try again due to ServerError
+
+            if error_attempts > SERVER_ERROR_RETRY:
+                # now we are past to what we would like to wait fail now
+                raise ProvisioningError(
```
netoarmando, line 555: Not related to your changes, but this exception is not being handled, so mrack execution exits here, I don't remember if this is expected. Output with forced 500 error: [mrack.log](https://github.com/neoave/mrack/files/6467171/mrack.log)

Tiboris: Oh so maybe we would like to handle it somehow...

Tiboris: feat: gracefully destroy servers after traceback

This allows us to create dummy Host object based
on the ProvisioningError traceback arguments
where we added also requirement which may have
caused the provisioning to fail so the worflow
can continue and remove already provisioned
servers after reprovision / abort strategy is
done so resources that may not be freed
to openstack will be returned.

src/mrack/providers/openstack.py
```diff
@@ -543,6 +543,7 @@ async def create_server(self, req):
         while True:
             try:
                 response = await self.nova.servers.create(server=specs)
+                # raise ServerError("a", "b")
```
Tiboris, line 546: Note for myself: this needs to be removed

src/mrack/providers/provider.py
```diff
@@ -28,6 +28,7 @@
 STRATEGY_RETRY = "retry"
 RET_CODE = 0
 HOST_OBJ = 1
+SPECS = 1
```
netoarmando, line 31: Could you please leave a comment describing this as the default index used to fetch data from the response or something like it?

Tiboris: Can do

netoarmando (approved):
@Tiboris thanks for this + addressing last minute issues LGTM

PR #121, opened 2021-06-08
fix: use BaseOS as variant for RHEL9.0 in Beaker

f-trivino:
Only BaseOS variant is available for RHEL9.0 in Beaker.

### DIFF
--- a/src/mrack/transformers/beaker.py
+++ b/src/mrack/transformers/beaker.py
@@ -45,7 +45,7 @@ class BeakerTransformer(Transformer):
         """Get variant for the host system to reqirement."""
         if "beaker_variant" in host:
             variant = host["beaker_variant"]
-        elif re.match(r"(rhel-8)", host["os"]):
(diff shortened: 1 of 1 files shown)


Tiboris:
We will need to use `[` to group 8 and 9`]`
```
In [1]: host8 = {"os": "rhel-8"}

In [2]: host9 = {"os": "rhel-9"}

In [3]: import re

In [4]: re.match(r"(rhel-8|9)", host8["os"])
Out[4]: <re.Match object; span=(0, 6), match='rhel-8'>

In [5]: re.match(r"(rhel-8|9)", host9["os"])

In [6]: re.match(r"(rhel-[8|9])", host9["os"])
Out[6]: <re.Match object; span=(0, 6), match='rhel-9'>

In [7]: re.match(r"(rhel-[8|9])", host8["os"])
Out[7]: <re.Match object; span=(0, 6), match='rhel-8'>
```

Tiboris (approved):
LGTM now!

PR #126, opened 2021-07-01
feat(Beaker): Add distro tag from provisioning conf

Tiboris:
Add feature which enables adding distro tags specified
in the provisioning config in beaker section like:
```
beaker:
    distro_tags:
        FEDORA-34%:
            - NIGHTLY-TESTED
```
This allows us to specify distros with custom tags.


netoarmando (approved):
LGMT. I understand that this is a "low-level" configuration, so `FEDORA-34%` instead of using the same "distro" key ('fedora-34') is fine.

Tiboris:
@netoarmando  yes i meant exactly like this, low level per distro which needs it.

PR #128, opened 2021-08-05
feat:  Update providers to have configurable strategy and minor fixes

Tiboris:
feat(Beaker): use timeout instead of number of retries
    using number of retries is being replaced by timeout
    to help users to redefine easily using provisioning config
    timeout variable should be more user friendly

    feat: add possibility to change strategy per provider in provisioning config
    change default openstack behavior to abort

    refactor(Beaker): Use hub url in messages

    refactor(Beaker): fix the typo in message

    refactor(static): remove method which is inherited

### DIFF
--- a/src/mrack/providers/beaker.py
+++ b/src/mrack/providers/beaker.py
@@ -255,8 +255,9 @@ chmod go-w /root /root/.ssh /root/.ssh/authorized_keys
 
             if prev_status != status:
                 logger.info(
-                    f"{self.dsp_name}: Job {beaker_id} has changed "
-                    f"status ({prev_status} -> {status})"
+                    f"{self.dsp_name}: Job "
+                    f"{self.hub._hub_url}jobs/"  # pylint: disable=protected-access
+                    f"{resource['id']} has changed status ({prev_status} -> {status})"
                 )
                 prev_status = status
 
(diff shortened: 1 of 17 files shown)


src/mrack/providers/beaker.py
```diff
@@ -250,7 +249,11 @@ async def wait_till_provisioned(self, resource):
         attempts = 0
         prev_status = ""
 
-        while attempts < self.max_attempts:
+        # let us use timeout variable which is in minutes to define
+        # maximum attempts of polling the beaker hub for job status
+        # with self.poll_sleep which defaults to 30s we should multiply
+        # this number with 2 to get the number of retries
```
netoarmando, line 255: This may cause troubles in the future. This commit is changing everything from `max_attemps` into `timeout`, but in the end the code is still counting the number of attempts in a not clear way.

Tiboris: What is your suggestion?

pvoborni: I agree with Armando. It may be very easy to cause problem in future if somebody just changes e.g. wait time without the "magic normalization".  I'd suggest to not mix units in one comparison. E.g. stop using attempts altogether and compare the time from the start of job with timeout.

For time-out case, you can use `else` block of `while else` construct. Instead of the current block which starts with `if attempts >= self.max_attempts:`.

pvoborni:
This MR also needs to contain documentation update as it is basically "hiding" some features of Mrack by disabling the retry strategy by default. There must be a way for people to learn what are all the possible strategies, what they do and what providers supports what. Resp. which strategy is supported by what providers.

Tiboris:
> This MR also needs to contain documentation update as it is basically "hiding" some features of Mrack by disabling the retry strategy by default. There must be a way for people to learn what are all the possible strategies, what they do and what providers supports what. Resp. which strategy is supported by what providers.

Noted i will update this asap

src/mrack/providers/beaker.py
Tiboris: I repushed and used timeout.

src/mrack/providers/podman.py
```diff
@@ -37,6 +37,7 @@ def __init__(self):
         """Initialize provider."""
         self._name = PROVISIONER_KEY
         self.dsp_name = "Podman"
+        self.max_attempts = 1 # for retry strategy
```
netoarmando: `max_attempts` or `max_retry`? I got confused what providers other than beaker are expecting on prov-config file.

Tiboris: yeah we need to change it to max_retry

netoarmando (approved):
LGMT, I verified various logs and also did a quick test on OpenStack to confirm it's not breaking that.

Issue #129, opened 2021-08-09
successfully provisioned resources are not freed when other providers fail


Tiboris:
e.g.:
```
2021-07-27T13:53:12     Beaker: Issuing provisioning of 5 host(s)
...
2021-07-27T13:53:12     OpenStack: Issuing provisioning of 6 host(s)
...
2021-07-27T13:53:12     OpenStack: Printing provisioned hosts

2021-07-27T13:53:12     OpenStack: active 63367b46-6e1f-4bf4-87db-a4778169b2d7 bkr-6-10.eagle.test 192.169.150.46 None None

2021-07-27T13:53:12     OpenStack: active fb388637-9bf0-43e4-b563-c9c474f9740c bkr-7-9.eagle.test 192.169.151.247 None None

2021-07-27T13:53:12     OpenStack: active 656988c8-a550-4818-a978-a0a7db2d0672 bkr-8-5.eagle.test 192.169.150.125 None None

2021-07-27T13:53:12     OpenStack: active 0c383fc0-2d4b-4136-a946-5f55dac05cae bkr-9-0.eagle.test 192.169.151.162 None None

2021-07-27T13:53:12     OpenStack: active 33bef762-152a-4851-baf6-862000feffcf osp-fedora-latest.eagle.test 192.169.151.33 None None

2021-07-27T13:53:12     OpenStack: active 09b6c825-1b8c-45f7-be9f-e4c857f7745e osp-win-2019.eagle.test 192.169.151.206 None None

...

2021-07-27T13:53:12     Beaker: All hosts reached provisioning final state (ACTIVE or ERROR)

2021-07-27T13:53:12     Beaker: Provisioning duration: 2:01:54.930595

...

2021-07-27T13:53:12     Beaker: SSH to host '192.169.34.116' timed out after 686.3s

2021-07-27T13:53:12     Beaker: Aborting provisioning due to error.

2021-07-27T13:53:12     Beaker: Error: New

2021-07-27T13:53:12     Beaker: Error: Could not establish ssh connection to host J:5636830 with IP 192.169.34.116

2021-07-27T13:53:12     Beaker: Given the error, will delete hosts

2021-07-27T13:53:12     Beaker: Issuing deletion

2021-07-27T13:53:12     Beaker: Deleting host by cancelling Job J:5636829

2021-07-27T13:53:12     Beaker: Deleting host by cancelling Job J:5636832

2021-07-27T13:53:12     Beaker: Deleting host by cancelling Job J:5636833

[...]

PR #130, opened 2021-08-09
fix: destroy active VMs after other providers fail

Tiboris:
fix(Virt): Add de-sync state and fix error parsing.

    fix: destroy active VMs after other providers fail

    Fixing the use case with multi provider failed run.
    In case of at least 2 providers in metadata and one
    failing, successfully provisioned hosts were not
    destroyed as part of freeing resources.
    e.g.:
    if:
    - Beaker fails
    - Openstack succeeds
    then
    - Beaker cancel jobs
    - Openstack resources keeps hanging not destroyed.
    Adding block of code to raise an exception when
    other than ProvisioningError occurred.
    
Resolves: https://github.com/neoave/mrack/issues/129

### DIFF
--- a/src/mrack/actions/up.py
+++ b/src/mrack/actions/up.py
@@ -17,10 +17,12 @@
 import asyncio
 import logging
 
-from mrack.errors import MetadataError
+from mrack.errors import MetadataError, ProvisioningError
 from mrack.providers import providers
 from mrack.transformers import transformers
 from mrack.utils import global_context, validate_dict_attrs
+
+PROVIDER_NAME_INDEX = 1
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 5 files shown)


src/mrack/actions/up.py
```diff
@@ -17,11 +17,13 @@
 import asyncio
 import logging
 
-from mrack.errors import MetadataError
+from mrack.errors import MetadataError, ProvisioningError
 from mrack.providers import providers
 from mrack.transformers import transformers
 from mrack.utils import global_context, validate_dict_attrs
 
+PROVIDER_NAME = 1
```
netoarmando, line 25: `PROVIDER_NAME_INDEX` maybe?

src/mrack/providers/provider.py
```diff
@@ -132,7 +132,7 @@ async def _wait_for_ssh(self, host, port=22, timeout=1200):
 
     async def _provision_base(
         self, reqs, res_check_timeout=60, res_busy_sleep=10
-    ):  # pylint: disable=too-many-locals, too-many-branches
+    ):  # pylint: disable=too-many-locals, too-many-branches, too-many-statements
```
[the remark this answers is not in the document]
Tiboris: what now :dagger:  :laughing:

src/mrack/actions/up.py
Tiboris: renamed

netoarmando (approved):

PR #131, opened 2021-08-11
feat: multiple fixes and features for Virt and fix the destroy success hosts logic

Tiboris:
feat: do not destroy if there are not success hosts
    
    renaming variable destroy to failed_providers
    to reflect what this variable really is.
    if there are no success hosts do not run
    branch with cleanup.
    
    chore: require testcloud >= 0.6.1
    
    feat(Virt): support testcloud v0.6.0 and later
    
    testcloud 0.6.1 now changes the invocation command
    add support to use testcloud ssh_path option.
    
    feat(Virt): log the tracebacks for Virt provider


src/mrack/providers/utils/testcloud.py
```diff
@@ -64,8 +64,9 @@ async def create(
         **kwargs,
     ):
         """Start a new testcloud instance."""
-        args = ["instance", "create", instance_name]
-        args.extend(["--url", image_url])
+        # print(instance_name)
```
pvoborni, line 67: leftover comment/print

src/mrack/providers/utils/testcloud.py
```diff
@@ -80,6 +81,11 @@ async def create(
             args.extend(["--no_graphics"])
         if kwargs.get("keep"):
             args.extend(["--keep"])
+        if kwargs.get("ssh_path"):
```
pvoborni, line 84: Isn't ssh_path used only for CoreOS instances?  AKA, does it help us somehow?

pvoborni (commented):
I just read the code. Mostly looks good. Have only minor comment and a question.

src/mrack/providers/utils/testcloud.py
Tiboris: oh right

src/mrack/providers/utils/testcloud.py
Tiboris: yes that is correct we just enabled support, unfortunately it would not help us

netoarmando (approved):
LGTM! Note: `testcloud` version `0.6.1` is available on F34.

PR #132, opened 2021-08-16
docs: fix the underline and verbosity of provider

Tiboris:
fix: make static provider more verbose
 

    docs: fix title underline too short

### DIFF
--- a/src/mrack/providers/static.py
+++ b/src/mrack/providers/static.py
@@ -14,11 +14,15 @@
 
 """Static provider."""
 
+import logging
+
 from mrack.errors import ValidationError
 from mrack.host import STATUS_ACTIVE
 from mrack.providers.provider import SPECS, STRATEGY_ABORT, Provider
 
 PROVISIONER_KEY = "static"
+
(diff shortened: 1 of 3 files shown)


netoarmando (approved):

src/mrack/providers/static.py
```diff
@@ -82,4 +87,5 @@ def prov_result_to_host_data(self, prov_result):
 
     async def delete_host(self, host_id):
         """Fake delete - pass but don't do anything."""
+        logger.info(f"{self.dsp_name}: Removing static host {host_id}")
```
pvoborni, line 90: Static host cannot be removed by mrack. Thus the log information needs to say something like: "Host {host_id} is static - not managed by Mrack, ignoring removal."

Tiboris: Will edit

pvoborni (approved):
ACKing  (provided that it works :) )

f-trivino:
it works also for me. I'm merging it.

PR #135, opened 2021-08-18
fix: attempts should be greater than max_retry

Tiboris:
number of attempts should be greater than max_retry
and not greater or equal than max_retry because
other way first attempt to provision will be
considered as first retry and provisioning will end
after reaching count 1 without reprovisioning retry
this was causing retry strategy to abort after
1st attempt even when max_retry was set to 1.


f-trivino (approved):
LGTM! tdudlak thanks for the patch

PR #136, opened 2021-08-20
feat(AWS): support provisioning of different arch

Tiboris:
AWS now support provisioning based on arch key from
host configuration (default: x86_64).
image is then selected from slightly modified format
for the aws provider from provisioning-config.yaml
where all the images are divided into groups
based on arch they are designed to be used.

### DIFF
--- a/src/mrack/data/provisioning-config.yaml
+++ b/src/mrack/data/provisioning-config.yaml
@@ -50,8 +50,9 @@ virt:
 
 aws:  # aws provider config
     images:
-        # list of ami images key is used in metadata ami image is used for provider e.g.
-        fedora-32: ami-0f1d60751946b66c5  # Fedora-Cloud-Base-32-1.6.x86_64-hvm-eu-central-1-gp2-0
+        # list of ami images based on arch (ami image used in metadata is overriding) e.g.
+        x86_64:
+            fedora-32: ami-0f1d60751946b66c5  # Fedora-Cloud-Base-32-1.6.x86_64-hvm-eu-central-1-gp2-0
 
 
     flavors:
--- a/src/mrack/transformers/aws.py
+++ b/src/mrack/transformers/aws.py
@@ -48,10 +48,15 @@ class AWSTransformer(Transformer):
 
     def create_host_requirement(self, host):
         """Create single input for AWS provisioner."""
-        required_image = host.get("image") or self._get_image(host["os"])
+        req_arch = host.get("arch", "x86_64")
+        arch_images = self.config.get("images")
+        req_image = host.get("image") or self._get_image(
+            host["os"], config_key=req_arch, provider_config=arch_images
+        )
+
         return {
             "name": host["name"],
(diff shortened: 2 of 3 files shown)


src/mrack/data/provisioning-config.yaml
```diff
@@ -50,8 +50,9 @@ virt:
 
 aws:  # aws provider config
     images:
-        # list of ami images key is used in metadata ami image is used for provider e.g.
-        fedora-32: ami-0f1d60751946b66c5  # Fedora-Cloud-Base-32-1.6.x86_64-hvm-eu-central-1-gp2-0
+        # list of ami images based on arch (ami image used in metadata is overriding) e.g.
+        x86_64:
+            fedora-32: ami-0f1d60751946b66c5  # Fedora-Cloud-Base-32-1.6.x86_64-hvm-eu-central-1-gp2-0
```
netoarmando, line 55: arch is not a required parameter if I got it right, so why not make it usable only by defining a new image name if needed? E.g.: `fedora-32-aarch64`.

Tiboris: Yes, this is possible approach and probably easier to maintain. However if we think from other side and metadata being reused for any provider IMO arch should not be bound to the `os` name, in this case fedora-32-aarch64.  This is probably question for @f-trivino  as well.

pvoborni (commented):
It feels kinda weird that image finding logic is not in `_get_image` method. It would be more natural to override the method in the provider and have the specific logic there. 

But I'd go even further. Same requirement might be needed for other providers, e.g. OpenStack. so why not create a common method which would handle all cases - override in host, default images and images for architecture. E.g. (not tested): 

```python
    def _get_image(self, host):
        """
        Get image name by OS name for host from provisioning config.

        Returns:
            1. image if specificied in host
            2. image by the os key
            3. default from the images if os is not found in keys
            4. image by the os key for architecture
            5. default from the image for architecture
            6. raises validation error
        """
        hostname = host["name"]
        os = host["os"]
        images = self.config["images"]
        
        # Try direct image first 
        image = host.get("image") 

        # Then non-arch specific images
        if not image: 
            image = get_config_value(images, os)
        
        if not image:
            arch = host.get("arch", "x86_64")
            arch_images = images.get(arch)
            if arch_images:
                image = get_config_value(arch_images, os)

        if not image:
            raise ValidationError(
                f"{self.dsp_name}: Image not found for {hostname}({os})"
            ) 
 
        logger.debug(
            f"{self.dsp_name}: Loaded image for {hostname}({os}): {image}"
        )
        return image
```

The downside is double meaning in provisioning config for something called as arch e.g., `x86_64` as if somebody woud define for host: `os: 'x86_64'` then it would pick the arch dict. 

But on the other side, if users don't care about arches then the `image` dict in provisioning config can remain simple. Also it is backwards compatible with current provisioning config.

Tiboris:
Lets close this for now

Issue #123, opened 2021-06-14
When beaker client returns Fault object it is not treated as error and runtime fails when unpacking values


Tiboris:
```
2021-06-08 12:42:43,991 mrack.providers.provider INFO Beaker: Preparing provider resources
2021-06-08 12:42:43,991 mrack.providers.provider INFO Beaker: Validating hosts definitions
2021-06-08 12:42:43,991 mrack.providers.provider INFO Beaker: Host definitions valid
2021-06-08 12:42:43,991 mrack.providers.provider INFO Beaker: Checking available resources
2021-06-08 12:42:43,991 mrack.providers.provider INFO Beaker: Resource availability: OK
2021-06-08 12:42:43,991 mrack.providers.provider INFO Beaker: Issuing provisioning of 1 host(s)
2021-06-08 12:42:43,991 mrack.providers.beaker INFO Beaker: Creating server
2021-06-08 12:42:46,749 mrack.providers.provider INFO Beaker: Provisioning issued
2021-06-08 12:42:46,749 mrack.providers.provider INFO Beaker: Waiting for all hosts to be active
2021-06-08 12:42:46,750 mrack.run ERROR cannot unpack non-iterable Fault object
Traceback (most recent call last):
  File "/home/tdudlak/Work/mrack/src/mrack/run.py", line 250, in handle
    ret_code = func(*args, **kwargs)
  File "/home/tdudlak/Work/mrack/src/mrack/run.py", line 274, in run
    mrackcli(obj={})  # pylint: disable=no-value-for-parameter,unexpected-keyword-arg
  File "/home/tdudlak/Work/mrack/venv/lib64/python3.9/site-packages/click/core.py", line 829, in __call__
    return self.main(*args, **kwargs)
  File "/home/tdudlak/Work/mrack/venv/lib64/python3.9/site-packages/click/core.py", line 782, in main
    rv = self.invoke(ctx)
  File "/home/tdudlak/Work/mrack/venv/lib64/python3.9/site-packages/click/core.py", line 1259, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/home/tdudlak/Work/mrack/venv/lib64/python3.9/site-packages/click/core.py", line 1066, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/home/tdudlak/Work/mrack/venv/lib64/python3.9/site-packages/click/core.py", line 610, in invoke
    return callback(*args, **kwargs)
[...]

PR #138, opened 2021-08-25
fix(Beaker): handle Fault exception when contacting hub

Tiboris:
Beaker provider may fail with throwing an Fault exception
when contacting beaker hub while creating requests.
We parse this exception string and behave as we can not
provision by creating fault beaker host object.

resolves: https://github.com/neoave/mrack/issues/123


Tiboris:
Commit fixing premature failure (fix(Beaker): delete_host handle premature deletion) without having proper host object from https://github.com/neoave/mrack/pull/137 will be needed to gracefully handle error.

PR #140, opened 2021-09-01
fix(AWS): Catch traceback when credentials are missing

Tiboris:
When we are missing export for AWS_CONFIG_FILE=...
boto throws an exception which we need to catch and
notify user nice way.


netoarmando (approved):
LGTM

PR #146, opened 2021-10-29
fix(OpenStack): Changes for AD hostname for the requirement

Tiboris:
fix(OpenStack): Use only first part of fqdn for host name

if the host name contains "." character which is not supported for
windows OS as host name or the host name is an FQDN we use only
first part of domain name as host identifier for openstack

refactor: use builtin operator to merge dictionary

mhcfg section from provisioning-config is default
and could be extended by user provided mhcfg section
from metadata, thus using this order for `|` operator

### DIFF
--- a/src/mrack/outputs/ansible_inventory.py
+++ b/src/mrack/outputs/ansible_inventory.py
@@ -120,7 +120,8 @@ class AnsibleInventoryOutput:
     def create_ansible_host(self, name):
         """Create host entry for Ansible inventory."""
         meta_host, meta_domain = get_host_from_metadata(self._metadata, name)
-        db_host = self._db.hosts[name]
+        host_name = meta_host["name"]
+        db_host = self._db.hosts[host_name]
 
         ip_addr = db_host.ip_addr
         ansible_host = resolve_hostname(ip_addr) or ip_addr
@@ -133,20 +134,20 @@ class AnsibleInventoryOutput:
         ansible_user = get_username(db_host, meta_host, self._config)
         password = get_password(db_host, meta_host, self._config)
         ssh_key = get_ssh_key(db_host, meta_host, self._config)
+        dom_name = meta_domain["name"]
 
         # Common attributes
         host_info = {
             "ansible_host": ansible_host,
             "ansible_python_interpreter": python,
             "ansible_user": ansible_user,
-            "meta_fqdn": name,
-            "meta_domain": meta_domain["name"],
+            "meta_fqdn": host_name,
+            "meta_host_name": name,
+            "meta_domain": dom_name,
             "meta_provider": db_host.provider.name,
             "meta_provider_id": db_host.host_id,
             "meta_ip": ip_addr,
-            "meta_dc_record": ",".join(
-                "DC=%s" % dc for dc in meta_domain["name"].split(".")
-            ),
+            "meta_dc_record": ",".join("DC=%s" % dc for dc in dom_name.split(".")),
         }
 
         if "restraint_id" in meta_host:
@@ -193,6 +194,7 @@ class AnsibleInventoryOutput:
 
         for host in provisioned.values():
             meta_host, _meta_domain = get_host_from_metadata(self._metadata, host.name)
+            host_name = meta_host["name"]  # FIXME here being leveraged
 
             # Groups can be defined in both "groups" and "group" variable.
             groups = meta_host.get("groups", [])
@@ -202,12 +204,12 @@ class AnsibleInventoryOutput:
 
             # Add only a reference custom groups
             for group in groups:
-                added = add_to_group(inventory, group, host.name)
+                added = add_to_group(inventory, group, host_name)
                 if not added:  # group doesn't exist
-                    add_group(inventory, group, host.name)
+                    add_group(inventory, group, host_name)
 
-            # Main record belongs in "all" group
-            all_group["hosts"][host.name] = self.create_ansible_host(host.name)
+            # Main record belongs in "all" group # FIXME using response
+            all_group["hosts"][host_name] = self.create_ansible_host(host.name)
         return inventory
 
     def create_output(self):
--- a/src/mrack/outputs/pytest_multihost.py
+++ b/src/mrack/outputs/pytest_multihost.py
@@ -80,6 +80,7 @@ class PytestMultihostOutput:
         for domain in mhcfg["domains"]:
             for host in domain["hosts"]:
                 provisioned_host = self._db.hosts.get(host["name"])
+
                 if not provisioned_host:
                     logger.error(f"Host {host['name']} not found in the database.")
                     continue
--- a/src/mrack/outputs/pytest_multihost.py
+++ b/src/mrack/outputs/pytest_multihost.py
@@ -70,12 +70,9 @@ class PytestMultihostOutput:
         ]
 
         # Use values from provisioning-config mhcfg section - as default values
-        for option in self._config["mhcfg"]:
-            # options defined in metadata takes precedence over options in
-            # provisioning-config as it is "closer" to user.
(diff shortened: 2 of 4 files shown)


src/mrack/outputs/pytest_multihost.py
```diff
@@ -70,16 +70,19 @@ def create_multihost_config(self):  # pylint: disable=too-many-branches
...
-            if option in mhcfg:
-                continue
-            mhcfg[option] = self._config["mhcfg"][option]
+        # options defined in metadata takes precedence over options in
+        # provisioning-config as it is "closer" to user.
+        mhcfg = self._config["mhcfg"] | mhcfg
 
         for domain in mhcfg["domains"]:
             for host in domain["hosts"]:
-                provisioned_host = self._db.hosts.get(host["name"])
```
pvoborni, line 82: This part and the part in util.py are not needed if we do not change the name of host in mrack db (other comment).

src/mrack/providers/openstack.py
```diff
@@ -549,6 +549,12 @@ async def create_server(self, req):
         if specs.get("flavor"):
             del specs["flavor"]
 
+        if "ad" in specs["group"] or "win" in specs["group"] or "win" in specs["image"]:
```
pvoborni, line 552: Let's use `is_windows_host` method from `util` to have consistent identification of windows hosts all over mrack.  Btw it works bit differently, so the Q is also: "how should we identify win hosts?":
```python
return (
        meta_host.get("os", "").startswith("win")
        or meta_host.get("os_type", "") == "windows"
    )
```

pvoborni (changes requested):
Hi, thanks for the PR.

 I think the commit should only do:
- use shortname in OpenStack server spec when provisioning windows hosts. (done)
- more `meta_` in ansible inventory (done)

Meaning:
- continue using provided name for non-windows (done)
- do not modify the host name in mrack DB (not done)

Not modifying the DB might be trickier given that how `to_host` and `prov_result_to_host_data` methods works. In OSP case, we are loosing the unmodified name.  https://github.com/neoave/mrack/pull/144/commits/1af33715a383e719749e424a38eaa6350fa3293d is creating some kind of preparation to solve this issue. Thus I'll continue reviewing that.

src/mrack/providers/openstack.py
Tiboris: Got it.

src/mrack/outputs/pytest_multihost.py
Tiboris: understood.

Tiboris:
Updated to use builtin `is_windows_host(host)`

src/mrack/outputs/ansible_inventory.py
```diff
@@ -206,8 +207,9 @@ def create_inventory(self):
                 if not added:  # group doesn't exist
                     add_group(inventory, group, host.name)
 
-            # Main record belongs in "all" group
-            all_group["hosts"][host.name] = self.create_ansible_host(host.name)
+            # Main record belongs in "all" group but use fqdn if possible
+            host_name = meta_host["name"] if is_windows_host(meta_host) else host.name
+            all_group["hosts"][host_name] = self.create_ansible_host(host.name)
```
pvoborni, line 212: As said in the previous review., we should be changing the behavior only in OpenStack vm name. Thus this should not be here.
AKA, keeping the exact value from metadata is good.

src/mrack/utils.py
```diff
@@ -165,17 +166,25 @@ def is_windows_host(meta_host):
...
     * os starts with 'win' or
-    * os_type is 'windows'
+    * os_type is 'windows' or
+    * group starts with 'WIN' or
+    * group uppercase starts with 'AD'
     """
     return (
         meta_host.get("os", "").startswith("win")
         or meta_host.get("os_type", "") == "windows"
+        or meta_host.get("group", "").upper().startswith("WIN")
```
pvoborni, line 176: Is the "WIN" and "AD" really needed here? I'm bit afraid that the check might trigger in correctly also on linux host where the group starts with "win" or "ad". E.g. "admin_machines" or "winners".  For that reason, the original check really does in on os name, or if somebody would have a Windows os named differently then there is the "os_type" to be explicit.

src/mrack/utils.py
```diff
@@ -153,7 +153,8 @@ def get_host_from_metadata(metadata, name):
     domains = metadata.get("domains", [])
     for domain in domains:
         for host in domain.get("hosts", []):
-            if host["name"] == name:
+            # use startswith as for ad we use only first part of FQDN
+            if host["name"].startswith(name):
```
pvoborni, line 157: Where is this change needed?

src/mrack/utils.py
```diff
@@ -165,17 +166,25 @@ def is_windows_host(meta_host):
...
         or meta_host.get("os_type", "") == "windows"
+        or meta_host.get("group", "").upper().startswith("WIN")
+        or meta_host.get("group", "").upper().startswith("AD")
     )
 
 
 def get_username(host, meta_host, config):
     """Find username from sources db/metadata/config."""
-    username = host.username or meta_host.get("username")
+    if host:
```
pvoborni, line 183: In what situation is this change needed? It seems to me that in all usages the host is defined.

pvoborni (changes requested):
I've highlighted the parts which were already mentioned in the previous review, but here it is more explicit.

src/mrack/utils.py
Tiboris: I agree that AD could be problematic, let me remove this.

src/mrack/utils.py
Tiboris: We are passing `host.name` as `name` parameter which is `Host` object attribute `name` and this is set based on response from server.
If we ask for name `ad.domain.com` the `host.name` will be `ad` as change is we ask for this name in OpenStack.

src/mrack/utils.py
Tiboris: This was probably one time error i will remove it

src/mrack/outputs/ansible_inventory.py
Tiboris: >We are passing `host.name` as `name` parameter which is `Host` object attribute `name` and this is set based on response from server.
If we ask for name `ad.domain.com` the `host.name` will be `ad` as change is we ask for this name in OpenStack.

Tiboris:
Hello @pvoborni  please have a look on #FIXME comments in code.

pvoborni (changes requested):
I don't really understand all the `FIXME` comments - they are not telling what to fix and reader is expected to somehow read the whole context, which is not possible as there is not indication that there is some.  Also the fact that there are 4 of them in a small commit tells that this PR should not be merged in this state.

Also the new changes are removing the agreed behavior to have shortname in Ansible inventory output. 

The actual patch is supposed to be quite small. The majority of work is actually in preparation to make it possible - as I wrote in the first review iteration. 

To illustrate what I mean, I've implemented it here: 
* the required change: https://github.com/neoave/mrack/pull/148/commits/a64b85608e3afc631af1a4074ad5e844a5001c60
* the Ansible inventory change: https://github.com/neoave/mrack/pull/148/commits/b35f9ff03fd0cfce58457105b3732fbfd7736c9a 
* the preration to have req available in to_host: https://github.com/neoave/mrack/pull/148/commits/89e3109281a1c168dd74ab7657ee4fb3d74d3879

Tiboris:
>  Also the fact that there are 4 of them in a small commit tells that this PR should not be merged in this state.

The FIXME comments were meant to show why and what change is needed and NOT to point to unfinished work or whatever else you have understood. Lets just say this is a communication issue and we misunderstood each other probably because my comments are not well written.

PR #148, opened 2021-11-22
feat: shortnames for windows machines in Openstack provider + shortname in Ansible inventory

pvoborni:
This is  an alternative implementation to https://github.com/neoave/mrack/pull/146 consisting of 3 changes. The actual fix, which is very small + 2 preparations where refactoring to pass the `req` is the biggest.

**fix(openstack): use shortnames for Windows vm names**
Windows machines get hostname often by Cloudbase-Init. But
windows host support setting only shortnames (domain is added
when the host joins AD domain). OpenStack derives the hostnane
from vm name (until nova version 2.90 where hostname attr is added).
Thus we set shortname to support this also on older OpenStacks.

**feat: add shortname in Ansible inventory output**
It is added in meta_hostname attribute as often shortnames are
referred as hostnames.

**refactor: pass req to to_host method**
Or in other words, do not loose it/return it from
wait_till_provisioned method.

That way a provider can loose all info about req
- even the host name and to_host will still have this
info accessible.

### DIFF
--- a/src/mrack/providers/openstack.py
+++ b/src/mrack/providers/openstack.py
@@ -34,7 +34,7 @@ from mrack.errors import (
 from mrack.host import STATUS_ACTIVE, STATUS_DELETED, STATUS_ERROR, STATUS_PROVISIONING
 from mrack.providers.provider import STRATEGY_ABORT, Provider
 from mrack.providers.utils.osapi import ExtraNovaClient, NeutronClient
-from mrack.utils import object2json
+from mrack.utils import get_shortname, is_windows_host, object2json
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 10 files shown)


src/mrack/providers/openstack.py
```diff
@@ -709,7 +714,7 @@ def prov_result_to_host_data(self, prov_result, req):
         result = {}
 
         result["id"] = prov_result.get("id")
-        result["name"] = prov_result.get("name")
```
Tiboris, line 712: Just a note, that this hides the actual name got/requested from OpenStack.

Tiboris (approved):
Thanks for the patches they do the work!

src/mrack/providers/openstack.py
pvoborni: Correct, but that's the goal, isn't it? It is more important to have in mrack db the actual name we use as a reference in all the parts of mrack. The name in the provider is not important - used for anything as we use provider ID for operations with the host in the provider.

PR #151, opened 2021-11-23
fix(OpenStack): Do not raise exception when using unavailable network

Tiboris:
In some cases when network has an outage or is temporarily
removed etc. mrack would raise an exception when getting
the "id" from network which is set to 'None' which is
obvious AttributeError: 'NoneType' object has no attribute 'get'

### DIFF
--- a/src/mrack/providers/openstack.py
+++ b/src/mrack/providers/openstack.py
@@ -206,6 +206,9 @@ class OpenStackProvider(Provider):
         networks = [self.get_network(net) for net in possible_networks]
         usable = []
         for network in networks:
+            if not network:
+                continue
+
             ips = self.get_ips(ref=network.get("id"))
 
             available = ips["total_ips"] - ips["used_ips"]
@@ -301,6 +304,13 @@ class OpenStackProvider(Provider):
         network = self.networks.get(name)
         if not network:
             network = self.networks_by_ref.get(ref)
+
+        if not network:
+            net_id = name or ref
+            logger.debug(
+                f"{self.dsp_name}: Failed to load network with name: '{net_id}'"
+            )
+
         return network
 
     def get_ips(self, name=None, ref=None):


src/mrack/providers/openstack.py
```diff
@@ -301,7 +309,8 @@ def get_network(self, name=None, ref=None):
         network = self.networks.get(name)
         if not network:
             network = self.networks_by_ref.get(ref)
-        return network
+
+        return network if network is not None else {"name": name}
```
pvoborni, line 313: As a user, I'd not expect this behavior from the `get_network` method. It's very "magicky" and thus I find it very problematic. Get method should return `None` if it doesn't exist, not some psedo object.

pvoborni (changes requested):
So the patch has 2 parts:
- warn if we want to work with network which is not available. A Q is whether we want it. As we are planning to have networks which will be expected to not be there thus it should not be a warning. It can be a debug level though.
- do not fail

So why not do it this way (untested):
```python
diff --git a/src/mrack/providers/openstack.py b/src/mrack/providers/openstack.py
index 1cdc491..f77e2cc 100644
--- a/src/mrack/providers/openstack.py
+++ b/src/mrack/providers/openstack.py
@@ -204,6 +204,7 @@ class OpenStackProvider(Provider):
         """
         possible_networks = self.network_pools[network_type]
         networks = [self.get_network(net) for net in possible_networks]
+        networks = [net for net in networks if net is not None]
         usable = []
         for network in networks:
             ips = self.get_ips(ref=network.get("id"))
@@ -301,6 +302,14 @@ class OpenStackProvider(Provider):
         network = self.networks.get(name)
         if not network:
             network = self.networks_by_ref.get(ref)
+
+        if not network:
+            id = name or ref
+            logger.debug(
+                f"{self.dsp_name}: Failed to load network"
+                f"with name or id: '{network.get(id)}'"
+            )
+
         return network
 
     def get_ips(self, name=None, ref=None):

```

src/mrack/providers/openstack.py
```diff
@@ -301,6 +304,13 @@ def get_network(self, name=None, ref=None):
         network = self.networks.get(name)
         if not network:
             network = self.networks_by_ref.get(ref)
+
+        if not network:
+            net_id = name or ref
+            logger.warning(
+                f"{self.dsp_name}: Failed to load network " f"with name: '{net_id}'"
```
pvoborni, line 311: There is `" f"` in the middle of the line.  

D you really want to use "warning" for this?

pvoborni (changes requested):

src/mrack/providers/openstack.py
Tiboris: black must have messed up thanks for the note

PR #154, opened 2021-11-29
fix(Openstack): raise Validation error when req has undefined network type

Tiboris:
when bogus network appear in metadata requirement
mrack continues and pass the network: None
to the openstack server as requirement thus it fails
This should fix the problems with networks that are
not defined/supported in provisioning-config.yaml

### DIFF
--- a/src/mrack/providers/openstack.py
+++ b/src/mrack/providers/openstack.py
@@ -188,7 +188,10 @@ class OpenStackProvider(Provider):
                 continue
             network_type = host.get("network")
             if not self._is_network_type(network_type):
-                continue
+                raise ValidationError(
+                    f"Network type '{network_type}' not found"
+                    f" in the provisioning config file"
+                )
 
             count = network_types.get(network_type, 0)
             count += 1
@@ -310,7 +313,6 @@ class OpenStackProvider(Provider):
             logger.debug(
                 f"{self.dsp_name}: Failed to load network with name: '{net_id}'"
             )
-
         return network
 
     def get_ips(self, name=None, ref=None):


src/mrack/providers/openstack.py
```diff
@@ -188,7 +188,10 @@ def _aggregate_networks(self, hosts):
                 continue
             network_type = host.get("network")
             if not self._is_network_type(network_type):
-                continue
+                raise ValidationError(
```
pvoborni, line 191: Don't remember it precisely from top of my head and would need to read the whole worfklow, but shouldn't we expect that also a specific network can be set here?  I.e. not only network type.

Tiboris: I believe it is set by `networks:` key I can recheck, i thought the same...

Tiboris: https://github.com/neoave/mrack/blob/cdb86cba3657bb0fa27d4b212f379d45af1a7447/src/mrack/providers/openstack.py#L184

pvoborni:
When I was looking deeper into this. I think that https://github.com/neoave/mrack/pull/117 introduced a regression where user cannot set network directly in metadata via: `networks: my_osp_network` entry. There reason is that move of the code caused that `names = host.get("networks")` on https://github.com/neoave/mrack/blob/master/src/mrack/providers/openstack.py#L186  will be always `None` as `networks` is not part of req. See https://github.com/neoave/mrack/blob/master/src/mrack/transformers/openstack.py#L65 It's because previously the `host` was host entry from metadata file. But in provider it is a `req`. 

Wrt this patch. This kind of validation should be done in `transformer` Probably on line 50 https://github.com/neoave/mrack/blob/master/src/mrack/transformers/openstack.py#L50

Tiboris:
I will have a look into that @pvoborni

Tiboris:
This patch is no longer valid.

PR #157, opened 2021-12-11
fix: SSH action - do not redirect to PIPEs

pvoborni:
SSH action was previously redirected to PIPEs and thus it stopped
working from a user perspective as nothing was visible.

This patch is bringing the original behavior back.

Regression was caused by commit: 19d513ba246ef4be02a0ce3d22fbb0faea971ce8


Tiboris:
perfect! thanks

PR #158, opened 2021-12-11
feat: search also provider config for username

pvoborni:
**feat: search also provider config for username**
    
Cloud images might have a different default user based on provider, e.g.
RHEL has ec2-user but OpenStack image has cloud-user. This allows to
define default user per OS in provider section of provisioning config.

**feat: find_value_in_config_hierarchy utility method**

For getting a configuration value from the most specific to the
least specific. Design in a way to be usable for multiple of attributes.

E.g. first looking in host object then host definition in metadata then
in provider configuration and then in global provisioning configuration.

### DIFF
--- a/src/mrack/utils.py
+++ b/src/mrack/utils.py
@@ -50,6 +50,71 @@ def get_config_value(config_dict, key, default=None):
     return val
 
 
+def get_value_or_dict_value(config_dict, attr, dict_name, key):
+    """
+    Look for a value in configuration dictionary.
+
+    Order of preference:
+    * dictionary 'dict_name' with key 'key'
+    * attribute: 'attr'
+    * None
+    """
+    if config_dict is None:
+        return None
+
+    value = None
+    value_dict = config_dict.get(dict_name)
+    if value_dict:
+        value = get_config_value(value_dict, key)
+        if value:
+            return value
+
+    value = config_dict.get(attr)
+    return value
+
+
+def find_value_in_config_hierarchy(
+    provisioning_config,
+    provider_key,
+    host,
+    meta_host,
+    attr,
+    dict_name,
+    key,
+    default=None,
+):
+    """
+    Look for first value in configuration hierarchy.
+
+    Order of preference:
+    * host value (attribute: attr)
+    * host metadata value (attribute: 'attr')
+    * provider config value (dictionary 'dict_name' with key 'key')
+    * provider default value (attribute: 'attr')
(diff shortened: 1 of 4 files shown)


pvoborni:
The new utility method can be used to a bit simplify other logic, e.g. as visible here: https://github.com/pvoborni/mrack/commit/1df617edbd4f970f6964c663da8514a88699c054  (could be sent when this is merged)

src/mrack/utils.py
```diff
@@ -198,13 +263,15 @@ def get_shortname(hostname):
...
     """Find username from sources db/metadata/config."""
-    username = host.username or meta_host.get("username")
-    default_user = get_config_value(config["users"], meta_host["os"])
-
+    provider = host.provider.name
+    dict_key = meta_host["os"]
+    default_user = None
     if is_windows_host(meta_host):
-        default_user = default_user or "Administrator"
+        default_user = "Administrator"
```
Tiboris, line 270: #NITPICK  I would add newline here so optically we can see if branch has only default user set but that change should be considered optional.
```suggestion
        default_user = "Administrator"
    
```

src/mrack/utils.py
```diff
@@ -198,13 +263,15 @@ def get_shortname(hostname):
 
 def get_username(host, meta_host, config):
     """Find username from sources db/metadata/config."""
-    username = host.username or meta_host.get("username")
-    default_user = get_config_value(config["users"], meta_host["os"])
-
+    provider = host.provider.name
+    dict_key = meta_host["os"]
```
Tiboris, line 267: I think we do not need this part and use the actual `host.provider.name` and `meta_host["os"]` in the function call. This would help me personally to not confuse and see more transparently what are the things we are passing to this function. As this change does not change the functionality i will leave it to you whether you change it or not.

Tiboris (approved):
Thanks for the patch @pvoborni I find the functionality very needed even when implementation is IMHO complex as we do not have mechanism for this yet.

Just to know what i have in mind its little bit confusing from the call perspective (because we do not have the mechanism...) that when we call 
```python
    username = find_value_in_config_hierarchy(
        config, provider, host, meta_host, "username", "users", dict_key, default_user
    )
```
the value we are looking for is in the middle of other input arguments.

I am approving the PR as we will benefit from this feature.

pvoborni:
Thanks for review. I agree with the suggested changes, but I'll merge as it is approved and I'm lazy to do them :) - could be done when this code is touched again.

Tiboris:
Agree

PR #159, opened 2021-12-14
feat: multiple aws enhancements (PrivateIPAddress, multiple security groups, Subnet)

pvoborni:
Requires: https://github.com/neoave/mrack/pull/158

**feat(aws): support for PrivateIpAddress**

When instance is provisioned with VPC which does not have a public
IP address, mrack is not able to get IP and fails. So changing the
behavior to try also Private IP if Public is not present.

**feat(aws): subnet support**

Adding possibility to define subnet to use for the instances. This
is needed, e.g. when instance needs to use a non-default VPC.

**feat(aws): multiple security groups**

Add support to be able to use multiple security groups and thus be able
to fine tune the inbount and outbound commnucation rules.

Note: there might be a small regression in validation in multiple security groups patch, where there is a possibility for alternate definitions but one is required and the other not. Not sure if we want to fix it - thus up to comments.

### DIFF
--- a/src/mrack/providers/aws.py
+++ b/src/mrack/providers/aws.py
@@ -40,7 +40,6 @@ class AWSProvider(Provider):
         self.dsp_name = "AWS"
         self.ami_ids: typing.List[str] = []
         self.ssh_key = None
-        self.sec_group = None
         self.instance_tags = None
         self.max_retry = 1  # for retry strategy
         self.status_map = {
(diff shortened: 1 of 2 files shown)


Tiboris:
Can we rebase?

Tiboris (approved):
Thanks for the patch @pvoborni LGTM

PR #160, opened 2021-12-15
feat(Openstack): poll openstack load when running can_provision()

Tiboris:
Can provision method was using only preloaded limits which caused
provisioning to fail always when resources at begining of the run
were already full. Now we reload limits with every can_provision()
method call and make sure load is refreshed and provisioning continues


pvoborni (approved):
Tried, seems to work, code looks also good.

PR #161, opened 2021-12-17
refactor: use hierarchy search for images and distros

pvoborni:
It may change some behavior a little but is mostly a refactoring.

This is unifying the logic to get image definitions across transformers.

WiIl use it as a base for a next patch in `aws` provider to get images by tag.


Tiboris (approved):
Works for me thanks for the patch!

PR #162, opened 2021-12-17
feat(aws): defining AMIs by tags

pvoborni:
rebased on top of: https://github.com/neoave/mrack/pull/161

Add a new way to define an AWS EC2 image in a provisiong config in addition to ami ID.

The following is valid:
```
images:
  fedora-34: ami-0d3c5199abb29a7ae
  fedora-35: { tag: { name: compose, value: mrack-fedora-35-latest } }
```

This allows for a seperate tool to upload new image with the same purpose and mrack automatically picks the latest one with the given tag.

Currently only one tag is supported and it searches in all images.

Possible improvements:
- more tags
- limiting to private images

Signed-off-by: Petr Vobornik <pvoborni@redhat.com>

PR #163, opened 2021-12-19
feat(aws): request spot instances

pvoborni:
This is only 1 commit but rebased on top of https://github.com/neoave/mrack/pull/162 (which is on top of https://github.com/neoave/mrack/pull/161 )

    
Spot instances are a great way how to save money for CI purposes.

https://aws.amazon.com/ec2/spot

This is inital implementation where the only possible option is to
ask for spot instance, but not define SpotOptions - this can be
extended later if wanted.

How to use:
- set `spot: True` in `aws` provider to enable it globally
- set `spot: True` in host section in job metadata file to request
  a spot instance for this particular host.

Additional changes:
- removes stopping instance on destroy as it doesn't work with
  default behaviour of spot instances + correctly handle exception when
  stopping fails (previous behaviour crashed mrack)
- prepares a parameters dict to be able to control what is passed to
  ec2.create_instances, + partly fixes a possible regression with
  SubNetId

### DIFF
--- a/src/mrack/providers/aws.py
+++ b/src/mrack/providers/aws.py
@@ -225,16 +225,24 @@ class AWSProvider(Provider):
         logger.info(f"{self.dsp_name}: Creating server")
         specs = deepcopy(req)  # work with own copy, do not modify the input
 
+        request = {
+            "ImageId": self.get_image(specs).image_id,
+            "MinCount": 1,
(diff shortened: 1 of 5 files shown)


Tiboris (approved):
This change is very welcome Thank you for the patch @pvoborni please rebase on master when dependent PR is merged.

PR #164, opened 2021-12-21
fix(Virt): remove password when provisioning windows

Tiboris:
testcloud is capable to inject password to linux
instances but not to windows one when provisioning
thus remove the password from the result when
the instance is identified as windows host.


pvoborni (approved):
Thanks for the PR.

I didn't test Windows use case, but it doesn't break non-windows and the patch looks good, thus ACK.

PR #165, opened 2021-12-21
feat: possibility to disable host DNS resolution in outputs

pvoborni:
Ansible inventory's `ansible_host` or pytest-multihost's `external_hostname` currently by default tries to resolve host's IP address and use the DNS name if resolution is successful.

This change allows to disable this behavior by setting:
   `resolve_host: False`
Somewhere in the configuration hierarchy (metadata host, provider config, or global provisioning config).


Tiboris (approved):
@pvoborni  thanks for the patch, works for me

PR #166, opened 2022-01-18
feat(aws): delete volumes on termination

pvoborni:
The behavior is default it can be disabled by setting:
```
  delete_volume_on_termination: false
```

In a config hierarchy of AWS provider.

### DIFF
--- a/src/mrack/data/provisioning-config.yaml
+++ b/src/mrack/data/provisioning-config.yaml
@@ -68,6 +68,7 @@ aws:  # aws provider config
         Name: "mrack-runner"
         mrack: "True"
         persistent: "False"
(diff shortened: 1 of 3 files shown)


Tiboris (approved):
Hello @pvoborni thanks for the patch, LGTM.

I would like to ask you if possible, please update example metadata:
https://github.com/neoave/mrack/blob/master/src/mrack/data/provisioning-config.yaml#L51-L70 
to contain latest stuff (image def by tags, delete_volume_on_termination)

pvoborni:
@Tiboris I'll leave merge on you as I don't know whether you want to review also the example provisioning config changes.

PR #167, opened 2022-03-21
feat: printout compose_id when using -latest image pointer for Openstack

dav-pascual:
In order to know what image is the "latest" tag pointing to in compose name in Openstack,
meta_compose_id and meta_compose_url fields from image meta information are
printed out in the log and added to ansible inventory output.

Implemented in a way that allows extension for other providers with extra custom fields in the future,
or images that don't have those fields at all.

### DIFF
--- a/src/mrack/host.py
+++ b/src/mrack/host.py
@@ -51,6 +51,7 @@ def host_from_json(host_data):
         host_data["username"],
         host_data["password"],
         host_data["error"],
+        host_data.get("meta_extra"),
     )
     return host
 
@@ -74,6 +75,7 @@ class Host:
         username=None,
         password=None,
         error_obj=None,
+        meta_extra=None,
     ):
         """Initialize host object."""
         self._provider = provider
@@ -87,6 +89,7 @@ class Host:
         self._password = password
         self._rawdata = rawdata
         self._error = error_obj
+        self._meta_extra = meta_extra
 
     def __str__(self):
         """Return string representation of host."""
@@ -120,6 +123,7 @@ class Host:
             "password": self._password,
             "rawdata": self._rawdata,
             "error": self._error,
+            "meta_extra": self._meta_extra,
         }
 
     @property
@@ -181,6 +185,11 @@ class Host:
     def password(self):
         """Get password for connecting to host."""
         return self._password
+
+    @property
+    def meta_extra(self):
+        """Get host extra meta information."""
+        return self._meta_extra
 
     async def delete(self):
         """Issue host deletion via associated provider."""
--- a/tests/unit/test_ansible_inventory.py
+++ b/tests/unit/test_ansible_inventory.py
@@ -67,6 +67,17 @@ def db(metadata):
     return get_db_from_metadata(metadata)
 
 
+@pytest.fixture()
+def db_meta_extra(metadata):
+    return get_db_from_metadata(
+        metadata,
+        host_extra={  # Sample data
+            "meta_compose_id": "ID.0-20220317.0",
+            "meta_compose_url": "http://dummy.com/compose/compose_id",
+        },
+    )
+
+
 def empty_layout():
     return {
         "all": {},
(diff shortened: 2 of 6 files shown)


dav-pascual:
As I stated, It's only implemented for Openstack images for now, and these values are optional since there are images that don't have meta_compose_id and meta_compose_url fields like Windows images. 

Compose Id and compose url are being printed out in mrack log, and mrack database has been extended to output them to ansible inventory file.

The extension of mrack database, consist of a dict in the host, which accepts such extra values. This allow us to use other values from other providers in the future.

tests/unit/test_ansible_inventory.py
```diff
@@ -118,3 +129,64 @@ def test_invalid_layouts(self, layout, db, metadata):
         with pytest.raises(ConfigError) as excinfo:
             ans_inv.create_inventory()
         assert "dictionary" in str(excinfo.value)
+
+    @pytest.mark.parametrize(
```
pvoborni, line 133: I don't think we need to parametrize this test on layout. Layout defines have groups are nested in each other this is not related to the tested feature - host attributes.

tests/unit/test_ansible_inventory.py
```diff
@@ -118,3 +129,64 @@ def test_invalid_layouts(self, layout, db, metadata):
...
+        first_host = inventory["all"]["hosts"][
+            metadata["domains"][0]["hosts"][0]["name"]
+        ]
+
+        assert (
+            "meta_compose_url" in first_host
+        ), "Host must have 'meta_compose_url' field"
+        assert "meta_compose_id" in first_host, "Host must have 'meta_compose_id' field"
+
+    @pytest.mark.parametrize(
```
pvoborni, line 160: Same comment as with the other parametrization.

tests/unit/test_ansible_inventory.py
```diff
@@ -118,3 +129,64 @@ def test_invalid_layouts(self, layout, db, metadata):
...
+            False,
+        ],
+    )
+    def test_meta_extra(self, layout, db_meta_extra, metadata):
+
+        config = provisioning_config(layout)
+        ans_inv = AnsibleInventoryOutput(config, db_meta_extra, metadata)
+        inventory = ans_inv.create_inventory()
+
+        first_host = inventory["all"]["hosts"][
```
pvoborni, line 151: think you can also use e.g.
```python
first_host = inventory["all"]["hosts"].values()[0]
```
which seems easier/more readable (untested though)

src/mrack/host.py
```diff
@@ -51,6 +51,7 @@ def host_from_json(host_data):
         host_data["username"],
         host_data["password"],
         host_data["error"],
+        host_data["meta_extra"],
```
pvoborni, line 54: This should use `host_data.get("meta_extra")` for the cases where e.g. `mrack list` was called with DB created by previous version of mrack. 

Now it causes e.g.:
```
  File "/home/pvoborni/.local/lib/python3.10/site-packages/mrack/dbdrivers/file.py", line 37, in __init__
    self.load()
  File "/home/pvoborni/.local/lib/python3.10/site-packages/mrack/dbdrivers/file.py", line 51, in load
    host = host_from_json(raw_host)
  File "/home/pvoborni/.local/lib/python3.10/site-packages/mrack/host.py", line 54, in host_from_json
    host_data["meta_extra"],
```

pvoborni (changes requested):
Hi, thanks for the patch. It mostly looks good. Have few smaller comments.  

I especially appreciate the comprehensive commit message and PR description.

tests/unit/test_ansible_inventory.py
dav-pascual: I also have observed that inventory output doesn't output host in the same order as in metadata.

Tiboris (approved):
thank you for the patch @dav-pascual  LGTM from my pov

tests/unit/test_ansible_inventory.py
pvoborni: It probably doesn't matter much.  All the host should get the attr - in current mock data preparation.  Anyway, this looks also fine.

dav-pascual:
Rebased with Black pre-commit hook patch

PR #169, opened 2022-03-29
chore: Update Black pre-commit hook version to 22.3.0 to solve click dependency issue

dav-pascual:
Recent release of Click makes current version of Black to break. This issue is solved in newest version of Black 22.3.0
See more about the issue in https://github.com/psf/black/issues/2964

Signed-off-by: David Pascual <davherna@redhat.com>

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -32,7 +32,7 @@ repos:
         files: ^src/.*
 
   - repo: https://github.com/ambv/black
-    rev: 20.8b1
+    rev: 22.3.0
     hooks:
       - id: black
         exclude: ^docs/.*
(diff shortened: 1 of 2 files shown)


Tiboris (changes requested):
Hehe looks like we missed something let us wonder a bit

Tiboris (approved):
Thanks for fixing the [test-requirements.txt](https://github.com/neoave/mrack/pull/169/files#diff-fac4c6890301d4de5c3f4266837803d5240c84a3d8b6c735bbc6a64c39d2f94e)

dav-pascual:
> Hehe looks like we missed something let us wonder a bit

solved :)

PR #171, opened 2022-04-04
fix: image transformer none value in requirements 

dav-pascual:
When OS image in metadata is not found in provisioning
config, there was a traceback causing mrack to fail.
Adding default value solves the problem and prints expected message.

### DIFF
--- a/src/mrack/transformers/transformer.py
+++ b/src/mrack/transformers/transformer.py
@@ -83,7 +83,9 @@ class Transformer:
     def _get_image(self, host):
         """Get image name by OS name from provisioning config."""
         operating_system = host["os"]
-        image = self._find_value(host, "image", "images", operating_system)
+        image = self._find_value(
+            host, "image", "images", operating_system, default=operating_system
+        )
 
         logger.debug(f"{self.dsp_name}: Found image {image} for {operating_system}")
         return image


Tiboris:
Hi @dav-pascual, thanks for the patch. Could you please paste before/after run log?

dav-pascual:
Before:

dav-pascual:
After fix:
```
OpenStack: Created requirement(s): [
    {
        "config_drive": true,
        "flavor": "ci.standard.medium",
        "group": "ipaserver",
        "image": "idm-rhel-8.5-ga-latest",
        "key_name": "idm-jenkins",
        "name": "r85.mrack.test",
        "network": "IPv4",
        "os": "rhel-8.5"
    },
    {
        "config_drive": true,
        "flavor": "ci.standard.medium",
        "group": "ipaserver",
        "image": "rhel-7.3",
        "key_name": "idm-jenkins",
        "name": "r73.mrack.test",
        "network": "IPv4",
        "os": "rhel-7.3"
    },
    {
        "config_drive": true,
        "flavor": "ci.standard.medium",
        "group": "ipaserver",
        "image": "idm-rhel-9.0-Server-nightly-latest",
        "key_name": "idm-jenkins",
        "name": "r9.mrack.test",
        "network": "IPv4",
        "os": "rhel-9.0"
    },
    {
        "config_drive": true,
        "flavor": "ci.standard.medium",
        "group": "ipaserver",
        "image": "idm-Fedora-Cloud-Base-33-latest",
        "key_name": "idm-jenkins",
        "name": "f33.mrack.test",
        "network": "IPv4",
        "os": "fedora-33"
    }
]
OpenStack: Preparing provider resources
OpenStack: Validating hosts definitions
OpenStack: Validating host: {
    "config_drive": true,
    "flavor": "ci.standard.medium",
    "group": "ipaserver",
    "image": "idm-rhel-8.5-ga-latest",
    "key_name": "idm-jenkins",
    "name": "r85.mrack.test",
    "network": "shared_net_7",
    "os": "rhel-8.5"
}
OpenStack: r85.mrack.test - OK
OpenStack: Validating host: {
    "config_drive": true,
    "flavor": "ci.standard.medium",
    "group": "ipaserver",
    "image": "rhel-7.3",
    "key_name": "idm-jenkins",
    "name": "r73.mrack.test",
    "network": "shared_net_7",
    "os": "rhel-7.3"
}
An unexpected exception occurred while provisioning
```

Tiboris (approved):
Thanks @dav-pascual works for me!

pvoborni:
This solution is better than before, but I don't think it is correct. The issue is in os-image mapping. I.e. there is no image defined for the os, e.g. "rhel-7.3".   Thus the error message should be something like: "Image name not found for os: rhel-7.3". 

`ERROR Image not found image: rhel-7.3`. Should be raised when there is no image with name `rhel-7.3` in OpenStack.

The validation error might be raised in the transformer itself as it is the place where the issue happens.

Tiboris:
@pvoborni Hi, i do not agree with your comment that it is not correct behavior. 
Before https://github.com/neoave/mrack/commit/c3a91ad9c36113ad14d4ebbab8985616a0a1359c commit the behavior was exactly the same and the idea was to use the 'os' as the fallback 'image' and try to provision like this (with for example pointing to exact ami/openstack image/distro etc.).
This was always something like hidden feature and the error was correctly (assuming the above behavior present in the releases up to https://github.com/neoave/mrack/releases/tag/v1.2.0 is the correct)

dav-pascual:
Image field is missing in the Hosts sub-entries table:
https://gitlab.cee.redhat.com/identity-management/idm-ci/-/blob/master/doc/pipeline-domains-specifications.md

Should we correct that?

Tiboris:
Definitely a good point to document this.

Tiboris:
slightly updating the docs in https://github.com/neoave/mrack/pull/174

PR #172, opened 2022-04-04
style: Increase readability of logs by using host

Tiboris:
Adding a host name from metadata or host object
where it is possible to increase readability of
the log messages and better debugging with only
using the log content. Before this patch there
were logging messages that were describing the
error but from the log message we could not tell
which host was causing that.

### DIFF
--- a/src/mrack/host.py
+++ b/src/mrack/host.py
@@ -193,6 +193,6 @@ class Host:
 
     async def delete(self):
         """Issue host deletion via associated provider."""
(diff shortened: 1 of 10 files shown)


Tiboris:
Yeah the thing is that we can not use it at every single place somewhere host object is ready somewhere only dict from requirement.

dav-pascual:
We could pass the pass the parameters just to build the string, but i think this is good enough!

Issue #173, opened 2022-04-05
AWS Provisioning failing when ec2 has some peeks in performance?


Tiboris:
Last days I have experienced some weird behavior in AWS provider and as I was exploring ec2 webUI i found out that instance has been in pending state which might (as i was too slow to get the ec2 instance details in webUI and i saw only this state) cause the mrack to fail later and finishing with this traceback:
```
2022-04-04T12:10:18     AWS Provisioning issued
2022-04-04T12:10:18     AWS Waiting for all hosts to be active
2022-04-04T12:10:18     An unexpected exception occurred while provisioning
2022-04-04T12:10:18     An unexpected exception occurred while provisioning
2022-04-04T12:10:18     An error occurred (InvalidInstanceID.NotFound) when calling the CreateTags operation: The instance ID 'i-0db1b3f72f0363db7' does not exist
2022-04-04T12:10:18     Traceback (most recent call last):
2022-04-04T12:10:18       File "/usr/local/lib/python3.9/site-packages/mrack/run.py", line 201, in handle
2022-04-04T12:10:18         ret_code = func(*args, **kwargs)
2022-04-04T12:10:18       File "/usr/local/lib/python3.9/site-packages/mrack/run.py", line 225, in run
2022-04-04T12:10:18         mrackcli(obj={})  # pylint: disable=no-value-for-parameter,unexpected-keyword-arg
2022-04-04T12:10:18       File "/usr/lib/python3.9/site-packages/click/core.py", line 829, in __call__
2022-04-04T12:10:18         return self.main(*args, **kwargs)
2022-04-04T12:10:18       File "/usr/lib/python3.9/site-packages/click/core.py", line 782, in main
2022-04-04T12:10:18         rv = self.invoke(ctx)
2022-04-04T12:10:18       File "/usr/lib/python3.9/site-packages/click/core.py", line 1259, in invoke
2022-04-04T12:10:18         return _process_result(sub_ctx.command.invoke(sub_ctx))
2022-04-04T12:10:18       File "/usr/lib/python3.9/site-packages/click/core.py", line 1066, in invoke
2022-04-04T12:10:18         return ctx.invoke(self.callback, **ctx.params)
2022-04-04T12:10:18       File "/usr/lib/python3.9/site-packages/click/core.py", line 610, in invoke
2022-04-04T12:10:18         return callback(*args, **kwargs)
2022-04-04T12:10:18       File "/usr/lib/python3.9/site-packages/click/decorators.py", line 21, in new_func
[...]

PR #175, opened 2022-05-04
feat(AWS): Move tagging into creation request itself

Tiboris:
Move tagging into creation request itself

When the creation of the instance is too slow
on the ec2 side we get a traceback that instance
does not exist (yet) and we need would need
to wait a bit in order to tag it.
Move tagging into request itself and do it at instance creation

adding some trivial patches:
- [refactor: remove collon from error string]
- [fix: return False when ValidationError is raised]

### DIFF
--- a/src/mrack/providers/aws.py
+++ b/src/mrack/providers/aws.py
@@ -196,7 +196,12 @@ class AWSProvider(Provider):
 
     async def prepare_provisioning(self, reqs):
         """Prepare provisioning."""
-        self.load_images(reqs)
+        try:
+            self.load_images(reqs)
+        except ValidationError as val_err:
+            logger.error(val_err)
+            return False
+
         return bool(reqs)
 
     async def validate_hosts(self, reqs):
--- a/src/mrack/providers/aws.py
+++ b/src/mrack/providers/aws.py
@@ -231,6 +231,12 @@ class AWSProvider(Provider):
         specs = deepcopy(req)  # work with own copy, do not modify the input
 
         del_vol = specs.get("delete_volume_on_termination", True)
+
+        # creating name for instance (visible in aws ec2 WebUI)
+        taglist = [{"Key": "name", "Value": req.get("name")}]
+        for key, value in self.instance_tags.items():
+            taglist.append({"Key": key, "Value": value})
+
         request = {
             "ImageId": self.get_image(specs).image_id,
             "MinCount": 1,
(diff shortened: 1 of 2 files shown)


Tiboris:
This could solve https://github.com/neoave/mrack/issues/173

pvoborni:
Hi, it will be cleaner and probably more maintainable to do the tagging after official wait. I.e. introduce e.g. in base provider a post-provision action. Create a 'tagging' method in the aws provider and call this method in and override of the post-provision-action method.

Tiboris:
Makes sense let me do it.

dav-pascual:
I found an alternative solution. You can set the tags in the moment of EC2 instance creation including the parameter `TagSpecifications` in **create_instances**
See reference https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/ec2.html

Tiboris:
I went with @dav-pascual solution for that. :)

pvoborni:
Yes, that's simpler and better solution.

src/mrack/providers/aws.py
```diff
@@ -226,6 +231,12 @@ async def create_server(self, req):
         specs = deepcopy(req)  # work with own copy, do not modify the input
 
         del_vol = specs.get("delete_volume_on_termination", True)
+
+        # creating name for instance (visible in aws ec2 WebUI)
+        taglist = [{"Key": "name", "Value": req.get("name")}]
+        for key in self.instance_tags:
+            taglist.append({"Key": key, "Value": self.instance_tags[key]})
```
dav-pascual, line 238: Just style, i think this way is more readable and better  for looping dicts k/v
```suggestion
        for key, value in self.instance_tags.items():
            taglist.append({"Key": key, "Value": value)
```

src/mrack/providers/aws.py
```diff
@@ -241,6 +252,7 @@ async def create_server(self, req):
                     },
                 },
             ],
+            "TagSpecifications": {"Tags": taglist},
```
dav-pascual, line 255: ResourceType would be 
```suggestion
            "TagSpecifications": {"ResourceType": "instance",
                                  "Tags": taglist},
```

dav-pascual (changes requested):

src/mrack/providers/aws.py
dav-pascual: I think it's builds actually failing because it needs that field

src/mrack/providers/aws.py
Tiboris: No prob i can update

dav-pascual:
Thanks @Tiboris 
Looks good to me.

dav-pascual (approved):
Thanks @Tiboris
Build tests passing.
Looks good to me.

PR #177, opened 2022-05-12
fix: Creating inventory with None host (re-running traceback)

dav-pascual:
The code in create_invetory method was looping hostnames from the DB,
which some of them may have already been deleted, which cause a None return
when trying to get their metadata (get_host_from_metadata).

This fixes traceback caused by rerunning mrack up with a different host name
without previously deleting mrackdb.json


dav-pascual:
The solution I came up with is (the simplest) is checking None return from get_host_from_metadata function. This should prevent further traceback, when not deleting mrackdb.json between reruns.

An alternative would be to check that the status of the host is set to active, or using success_host list.

Tiboris (approved):
Thanks for the patch @dav-pascual lets merge it.

PR #180, opened 2022-07-19
feat: copyign meta_ attributes from host to ansible inventory

pvoborni:
Make it possible to define arbitratry attributes prefix with meta_
in host section of job metadata file.

All of these attributes are then copied to host part in ansible
inventory output.

It can serve for creation of playbooks which consumes these attributes.

Addtional, but probably rare use case, is to override the meta_ attrs
which are derived from other others by mrack.

### DIFF
--- a/tests/unit/test_ansible_inventory.py
+++ b/tests/unit/test_ansible_inventory.py
@@ -190,4 +190,40 @@ class TestAnsibleInventory:
         assert "meta_something_else" not in srv2
         assert "meta_readonly_dc" in srv2
         assert srv2["meta_readonly_dc"] == "no"
-        assert srv2["meta_os"] == "windows-2022"
+        assert srv2["meta_os"] == "fedora-32"
+
+    def test_arbitrary_attrs(self):
+        """
+        Test that values defined in `ansible_inventory` dictionary in host part
+        of job metadata file gets into host attributes in generated ansible
+        inventory.
+        """
+        metadata = metadata_extra()
+        m_srv1 = metadata["domains"][0]["hosts"][0]
+        m_srv2 = metadata["domains"][0]["hosts"][1]
+        m_srv1["ansible_inventory"] = {
+            "readonly_dc": "yes",
+            "something_else": "for_fun",
+        }
+        m_srv2["ansible_inventory"] = {
+            "no_ca": "yes",
+            "something_else": "for_fun",
+        }
+
+        config = provisioning_config()
+        db = get_db_from_metadata(metadata)
+        ans_inv = AnsibleInventoryOutput(config, db, metadata)
+        inventory = ans_inv.create_inventory()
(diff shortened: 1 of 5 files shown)


tests/unit/test_ansible_inventory.py
```diff
@@ -84,6 +84,37 @@ def empty_layout():
     }
 
 
+@pytest.fixture()
+def metadata_extra():
+    hosts = [
+        {
+            "name": "srv1.example.test",
```
kaleemsiddiqu, line 91: Can we replace one data set without window os and with some other custom meta_* param and without meta_readonly_dc param? to just have different data to verify other non window and non meta_readonly_dc param data.

pvoborni:
I've added also the remaining features + to fixes to address some crashes. I only unit-tested the PR.

**feat(ansible-inventory): host arbitrary attributes**

Adding a possibility to add arbitrary attributes to host part of
generated Ansible inventory via defining them per-host in new dictonary
`ansible_inventory`.

This is useful for creating more complex jobs which are using ansible
playbooks or roles requiring to get vars from hosts.

**fix(pytest-multihost): crash when group is not defined**

It should possible to define host groups either via "group" or "groups"
variable. But pytest-multihost output was crashing when it was done via
"groups".

This fix makes it not to crash if both are missing. It also makes it
possible to use first group from "groups" if groups are defined -
similar as in Ansible inventory output.

**fix(pytest-multihost): crash when mhcfg is missing in prov. config**

`mhcfg` is not required key in provisinong config, but pytest-multihost
plugin crashes if it is missing. This fix makes it not crash.


**feat(pytest-multihost): arbitrary attributes for hosts**

Make it possible to add arbitrary attributes per-host via defining them
in pytest_multihost dictionary in host section of job metadata file.

tests/unit/test_ansible_inventory.py
pvoborni: Some of the uses cases with not having the values present are tested by another tests in the `test_ansible_inventory` output. 

Changed one of the hosts to not be a windows host.

Tiboris:
I have checked the code and it look good to me. The small nitpick as we decided to go with the `_` underscore the commit messages are well with `-` (`feat(ansible-inventory)`) i do not mind but if you are wiling to rewrite and it makes sense to you as well please rewrite it. 

I will test this and the bugs you found and provide feedback !

Tiboris (approved):
Thanks works for me!

pvoborni:
For the commit message summary lines, e.g.,  `feat(ansible-inventory`. It should not reflect the dictionary name. It supposed to refer to the whole ansible inventory output - as a component. But I think we did not agree on any specific names of the mrack components. Thus I'd leave it be.

Tiboris:
I totally agree :)

PR #183, opened 2022-09-07
fix: Beaker log polling to logfile instead of console

Tiboris:
Beaker was too chatty in console output so we moved the messages to the debug level so it is logged in the mrack.log only. Added debug message to the
function responsible for polling beaker for job status.


pvoborni (approved):
Hi,

Looks good to me, works fine. 

Thanks for the patch.

Tiboris:
Thanks for the review, merging

PR #184, opened 2022-09-07
feat(Beaker): Support custom configurable ks_meta values

Tiboris:
get the ks_meta from host configuration, global
provisioning config or default hardcoded in code.
The priority is following:
 - host
 - provisioning-config.yaml
 - default (if nothing set)
https://beaker-project.org/docs/user-guide/customizing-installation.html#kickstart-metadata-ks-meta

### DIFF
--- a/src/mrack/transformers/beaker.py
+++ b/src/mrack/transformers/beaker.py
@@ -66,6 +66,25 @@ class BeakerTransformer(Transformer):
 
         return (required_distro, variant)
 
+    def _get_ks_meta(self, host):
+        """
+        Get `ks_meta` value from host or provisioning config or default if not defined.
+
+        The priority is following:
+            - host
+            - provisioning-config.yaml
+            - default from provisioning config
+            - empty if not defined in provisioning config
+
+        """
+        res = self._find_value(
+            host.get(CONFIG_KEY, {}),
+            "ks_meta",
+            "kickstart_metadata",
+            host["os"],
+        )
+        return res
+
     def create_host_requirement(self, host):
         """Create single input for Beaker provisioner."""
         distro, variant = self._get_distro_and_variant(host)
@@ -77,5 +96,6 @@ class BeakerTransformer(Transformer):
             "meta_distro": "distro" in host,
             "arch": host.get("arch", "x86_64"),
             "variant": variant,
+            "ks_meta": self._get_ks_meta(host),
             f"mrack_{CONFIG_KEY}": host.get(CONFIG_KEY, {}),
         }
--- a/tests/unit/mock_data.py
+++ b/tests/unit/mock_data.py
@@ -207,7 +207,11 @@ def provisioning_config(inventory_layout=None):
             },
             "pubkey": "config/id_rsa.pub",
             "reserve_duration": 86400,
-            "max_attempts": 240,
+            "kickstart_metadata": {
+                "default": "PROV_CONF_DEFAULT",
+                "rhel-8.6": "PROV_CONF_RHEL86_KS_META",
+                "c9s": "PROV_CONF_CENTOS_KS_META",
+            },
             "timeout": 120,
         },
         "users": {
--- a/tests/unit/test_beaker_transformer.py
+++ b/tests/unit/test_beaker_transformer.py
@@ -1,6 +1,5 @@
 import pytest
 
-from mrack.context import global_context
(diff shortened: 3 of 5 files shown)


src/mrack/transformers/beaker.py
```diff
@@ -15,6 +15,7 @@
 """Beaker transformer module."""
 import re
 
+from mrack.context import global_context
```
pvoborni, line 18: Global context should not be needed here to access provisioning config. Transformer should have it in `self.config` (provider) or `self._config` (global). I'd even say that using global context is an anti-pattern.

tests/unit/mock_data.py
```diff
@@ -207,7 +207,7 @@ def provisioning_config(inventory_layout=None):
             },
             "pubkey": "config/id_rsa.pub",
             "reserve_duration": 86400,
-            "max_attempts": 240,
+            "ks_meta": "GLOBAL_KS_META",
```
pvoborni, line 210: Wouldn't `ks_meta` be different per os? I.e. shouldn't we rather have it as a dict with OS keys and with `default` as a fallback?

src/mrack/transformers/beaker.py
```diff
@@ -66,6 +67,24 @@ def _get_distro_and_variant(self, host):
...
+        """
+        Get `ks_meta` value from host or provisioning config or default if not defined.
+
+        The priority is following:
+            - host
+            - provisioning-config.yaml
+            - default (if nothing set)
+
+        """
+        return host.get(CONFIG_KEY, {}).get(
```
pvoborni, line 80: Wouldn't `transformer._find_value` or some direct usage of `find_value_in_config_hierarchy` be more useful here? Especially if we use per-os config in provisioning config (comment below).

pvoborni (changes requested):

tests/unit/mock_data.py
Tiboris: I can rewrite it

src/mrack/transformers/beaker.py
Tiboris: removed in future commit

src/mrack/transformers/beaker.py
Tiboris: let me try

src/mrack/transformers/beaker.py
```diff
@@ -66,6 +66,26 @@ def _get_distro_and_variant(self, host):
...
+            - default from provisioning config
+            - empty if not defined in provisioning config
+
+        """
+        res = self._find_value(
+            host.get(CONFIG_KEY, {}),
+            "ks_meta",
+            "kickstart_metadata",
+            host["os"],
+            default=self.config.get("kickstart_metadata", {}).get("default", ""),
```
pvoborni, line 85: Defining the default this way should not be needed. This behavior is build-in  by methods called indirectly by find_value_in_config_hierarchy . Here the `default` should mean, use it if nothing else work.

tests/unit/test_beaker_transformer.py
```diff
@@ -1,5 +1,6 @@
 import pytest
 
+from mrack.context import global_context
```
pvoborni, line 3: Shouldn't this be removed? I don't think it is needed. Or is it?

pvoborni (changes requested):
I see that github also shows some pre-commit linter issue  in the code part - this is good.

tests/unit/test_beaker_transformer.py
Tiboris: you are correct thanks!

src/mrack/transformers/beaker.py
Tiboris: removed, I forgot.

pvoborni (approved):

pvoborni:
Approved. It works. Looks OK. But have a Q: Why is one key named `ks_meta` and the other `kickstart_metadata` shouldn't it be the same, or is there some other specific reason?

Tiboris:
To reply to above comment the reason is just that it is more readable in my opinion ks_meta is variable passed to provider which is supported to it but it is basically kickstart_metadata which user can pass as a string. If that's not what we want we can rename it to ks_meta of course.

pvoborni:
We can merge this. My only worry is that it will confuse users. Time will tell.

PR #185, opened 2022-09-12
feat(Beaker): support configurable jobxml specs 

Tiboris:
Moving the translation of requirement to specs
to the beaker transformer and support more options. Now the beaker specific configurations
can be defined in provisioning config,
host's beaker section. If nothing is specified
sane defaults are configured.

### DIFF
--- a/src/mrack/transformers/beaker.py
+++ b/src/mrack/transformers/beaker.py
@@ -90,8 +90,7 @@ class BeakerTransformer(Transformer):
             return []
 
         res_ks_append = ["%post"]
-
-        if pubkeys is not None:
+        if pubkeys:
             ks_append += self._allow_ssh_keys(pubkeys)
 
         res_ks_append += ks_append
--- a/src/mrack/transformers/beaker.py
+++ b/src/mrack/transformers/beaker.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 """Beaker transformer module."""
+import os
 import re
 
 from mrack.providers.provider import STRATEGY_ABORT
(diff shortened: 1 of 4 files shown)


Tiboris:
Added one more commit.

Tiboris:
I need to fix the tests for the use cases.

src/mrack/transformers/beaker.py
```diff
@@ -97,5 +147,60 @@ def create_host_requirement(self, host):
...
+                "tasks",
+                host["os"],
+                default=[  # we use dummy task because beaker require a task in recipe
+                    {"name": "/distribution/dummy", "role": "STANDALONE"}
+                ],
+            ),
             f"mrack_{CONFIG_KEY}": host.get(CONFIG_KEY, {}),
         }
+
+        specs.update(
```
pvoborni, line 191: Trivial (no need to change the PR if nothing else is changed):  Is there a need for `spec.update` ? It looks that it can be part of the dict definition.

pvoborni (changes requested):
The code looks good to me, but it would be good to cover the changed logic with unit tests, this is a technical debt from the past which should be fixed.

src/mrack/transformers/beaker.py
Tiboris: yes it can let me update

Tiboris: Updated and commit with tests added

pvoborni (approved):
I've tested this locally, it seems to work. The code also looks OK. Thus approving.

That said, it would be super to have unit tests also for the other parts touched in this MR, namely:  `task`, `retention_tag`, `product`  a ssh part of `ks_append`. But these can be added separately.

Tiboris:
Thanks for the review, Created issue for missing tests

PR #187, opened 2022-09-19
refactor(beaker): move distro_tags and hostRequires to transformer

pvoborni:
**refactor(beaker): move distro_tags and hostRequires to transformer**

From provider as transformer is the component which should work with
provisioning config.

Also remove the mrack_beaker from specs as it is no longer needed.

**fix: issue when searching for value when dict_name == attr**

When searching for value using find_value_in_config_hierarchy the method
might return the entire dictionary with name according to dict_name when
the search attribute name is the same as dict_name and the dict doesn't have
a default value nor value according to provided key.


Tiboris (approved):
Thanks for the patch @pvoborni I have tried this manually and and in CI works fine.

Issue #113, opened 2021-04-30
Consider dividing mrack to multiple rpms 


Tiboris:
Splitting the spec file to have packages like mrack, mrack-all, mrack-openstack, mrack-virt, mrack-containers, mrack-aws, ... + associated changes to handle missing python deps

PR #190, opened 2022-09-30
feat: Split mrack into multiple rpms

Tiboris:
Splitting mrack to:
    - mrack
    - mrack-cli
    - python3-mracklib
    - python3-mrack-aws
    - python3-mrack-beaker
    - python3-mrack-openstack
    - python3-mrack-podman
    - python3-mrack-virt
   
    where python3-mracklib contains only static provider and all the
    scripts and provisioning base logic.
    we can install mrack-{aws,beaker,openstack,podman,virt}
    providers in addition to static provider extending
    the functionality and create modular solution where only
    needed providers can be installed.
    Package mrack contains all supported providers.
    Package mrack-cli contains mrack command.
    We are using Suggests directives for OpenStack dependencies
    so in fedora where these are not available build will pass.

Resolves: https://github.com/neoave/mrack/issues/113 

This MR depends on:
- [x] https://github.com/neoave/mrack/pull/191 
- [x] https://github.com/neoave/mrack/pull/194 
- [x] discuss the pacakge division

### DIFF
--- a/README.md
+++ b/README.md
@@ -121,9 +121,36 @@ pip install mrack
 It is also available for Fedora 32+ via [COPR](https://copr.fedorainfracloud.org/coprs/g/freeipa/neoave/package/mrack/):
 
 ```
+# enable the copr at your system
 sudo dnf copr enable @freeipa/neoave
+
+# install mrack package with all supported providers and cli
 sudo dnf install mrack
+
+# install mrack command and e.g. AWS provider
+sudo dnf install mrack-cli python3-mrack-aws
+
+# install mrack library to be later used as an import with support of OpenStack provider:
+sudo dnf install python3-mrack-openstack
+
 ```
+---
+**NOTE**
+
+Above command will install `mrack` package which depends on (contains) following packages:
+- mrack-cli
+- python3-mracklib
+- python3-mrack-aws
+- python3-mrack-beaker
+- python3-mrack-openstack
+- python3-mrack-podman
+- python3-mrack-virt
+
+The `python3-mracklib` is mandatory as it contains core mrack library.
+To install only some of `mrack` providers you should install `python3-mrack-{provider_name}`.
+To install mrack command a package `mrack-cli` is required.
+
+---
 
 ## Run mrack
 
--- a/mrack.spec
+++ b/mrack.spec
@@ -1,46 +1,103 @@
 # Created by pyp2rpm-3.3.5
-%global srcname mrack
-
-Name:           %{srcname}
+Name:           mrack
 Version:        1.8.1
 Release:        1%{?dist}
 Summary:        Multicloud use-case based multihost async provisioner for CIs and testing during development
 
 License:        Apache License 2.0
 URL:            https://github.com/neoave/mrack
-Source0:        https://github.com/neoave/mrack/releases/download/v%{version}/mrack-%{version}.tar.gz
+Source0:        https://github.com/neoave/mrack/releases/download/v%{version}/%{name}-%{version}.tar.gz
 BuildArch:      noarch
 
 BuildRequires:  python3-devel
-BuildRequires:  python3-AsyncOpenStackClient
-BuildRequires:  beaker-client
-BuildRequires:  python3-boto3
-BuildRequires:  python3-botocore
 BuildRequires:  python3-click
 BuildRequires:  python3-pyyaml
 BuildRequires:  python3-setuptools
 
+# coma separated list of provider plugins
+%global provider_plugins aws,beaker,openstack,podman,virt
+
+Requires:       %{name}-cli == %{version}-%{release}
+Requires:       python3-%{name}lib == %{version}-%{release}
+Requires:       python3-%{name}-aws == %{version}-%{release}
+Requires:       python3-%{name}-beaker == %{version}-%{release}
+Requires:       python3-%{name}-openstack == %{version}-%{release}
+Requires:       python3-%{name}-podman == %{version}-%{release}
+Requires:       python3-%{name}-virt == %{version}-%{release}
+
+# We filter out the asyncopenstackclient dependency of this package
+# so it is not forcing installation of missing dependencies in Fedora
+# Once python3-AsyncOpenStackClient is in fedora we can drop this line
+%global __requires_exclude asyncopenstackclient
+
 %description
-mrack is a provisioning library for CI and local multi-host testing supporting multiple
-provisioning providers (e.g. OpenStack, Beaker). But in comparison to other multi-cloud
-libraries, the aim is to be able to describe host from application perspective.
+mrack is a provisioning tool and a library for CI and local multi-host testing supporting
+multiple provisioning providers (e.g. AWS, Beaker, Openstack). But in comparison to other
+multi-cloud libraries, the aim is to be able to describe host from application perspective.
 
-%{?python_provide:%python_provide %{srcname}}
+%{?python_provide:%python_provide %{name}}
 
-Requires:       python3-AsyncOpenStackClient
-Requires:       beaker-client
-Requires:       podman
-Requires:       python3-boto3
-Requires:       python3-botocore
+%package        cli
+Summary:        mrack command
+Requires:       python3-%{name}lib == %{version}-%{release}
+
+%package -n     python3-%{name}lib
+Summary:        Core mrack libraries
 Requires:       python3-click
 Requires:       python3-pyyaml
-Requires:       testcloud
 Requires:       sshpass
 
+%package -n     python3-%{name}-aws
+Summary:        AWS provider plugin for mrack
+Requires:       python3-%{name}lib == %{version}-%{release}
+Requires:       python3-boto3
+Requires:       python3-botocore
+
+%package -n     python3-%{name}-beaker
+Summary:        Beaker provider plugin for mrack
+Requires:       python3-%{name}lib == %{version}-%{release}
+Requires:       beaker-client
+
+%package -n     python3-%{name}-openstack
+Summary:        Openstack provider plugin for mrack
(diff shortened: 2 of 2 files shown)


mrack.spec
```diff
@@ -1,5 +1,5 @@
 # Created by pyp2rpm-3.3.5
-%global srcname mrack
```
pvoborni, line 2: typo: `pyton3`

pvoborni: And actually, python applications doesn't have to start with `python3`. 
As https://docs.fedoraproject.org/en-US/packaging-guidelines/Python/#_application_naming says:

> Packages that primarily provide applications, services or any kind of executables SHOULD be named according to the general [Fedora naming guidelines](https://docs.fedoraproject.org/en-US/packaging-guidelines/Naming/) (e.g. ansible).

Probably the most correct thing would be to have packages:

- mrack (packaging mrack CLI)
- python3-mrack (core of the mrack - the library to be imported e.g. in tmt)
- python3-mrack-openstack (the provider)

pvoborni (commented):
I didn't test, so these are just my initial comments after reading the PR. It mostly looks good to me.

mrack.spec
Tiboris: I have re-pushed without the patch adding the python- prefix.
We should decide how to divide the package itself i have pushed this PR to initiate discussion basically as we need to agree on something which makes sense for us as well.

My question is, should we create package per each provider from (AWS, Virt, Beaker, Openstack, Podman) and keep static as a base for the library or should we make 

```
└─ mrack   
       └─  python3-mrack-all (maybe without this part)
                     └─ python3-mrack-aws
                     │          └─  python3-mrack (Static)  
                     ├─ python3-mrack-beaker
                     │          └─  python3-mrack (Static)  
                     ├─ python3-mrack-openstack
                     │          └─  python3-mrack (Static)  
                     ├─ python3-mrack-podman
                     │          └─  python3-mrack (Static)  
                     └─ python3-mrack-virt
                                └─  python3-mrack (Static)  

```

Tiboris: How would you divide the package?

Tiboris:
@praiskup Would you mind having a look here?

mrack.spec
```diff
@@ -12,48 +12,120 @@ Source0:        https://github.com/neoave/mrack/releases/download/v%{version}/mr
...
 
-Requires:       python3-AsyncOpenStackClient
-Requires:       beaker-client
-Requires:       podman
-Requires:       python3-boto3
-Requires:       python3-botocore
+Requires:       %{name}-core
+Requires:       %{name}-aws
+Requires:       %{name}-beaker
+Suggests:       %{name}-openstack
```
praiskup, line 32: This isn't needed IMO, the `mrack-openstack` sub-package is installable because *that* package uses "only" Suggests for problematic deps.

mrack.spec
```diff
@@ -1,7 +1,7 @@
 # Created by pyp2rpm-3.3.5
 %global srcname mrack
 
-Name:           %{srcname}
+Name:           mrack
```
praiskup, line 4: A nit ... you could even drop the %srcname definition now?  I don't think it was necessary to move from `srcname` to just `name`, but admittedly the spec looks prettier now.

Tiboris: Thanks! Yeah I forgot to remove the `srcname`

mrack.spec
Tiboris: So at this level `Requires` is safe correct?

praiskup:
So previously there was just `mrack`, and also the `mrack` package stays here ... serving the same purpose as before, bringing in all the new deps.  So you don't have to do any artificial dependency "obsoletes" hacks or so, good (you just update to an updated mrack.rpm, and all the deps will be installed automatically).  Overall it looks fine to me.

Tiboris:
@pvoborni do we want the mrack-cli package? it would contain mrack script only

Tiboris:
Right now the packages should reflect this dependency graph:
```
└─             mrack   
                     └─  mrack-core (Static)    
                     └─ mrack-aws
                     │     └─  mrack-core (Static)    
                     ├─ mrack-beaker
                     │     └─  mrack-core (Static)    
                     ├─ mrack-openstack
                     │      └─  mrack-core (Static)   
                     ├─ mrack-podman
                     │      └─  mrack-core (Static)   
                     └─mrack-virt
                               └─  mrack-core (Static)  

```

praiskup:
> mrack-all (virtual package which requires all providers in this repo packaged for given OS)

This sounds doable?  Not the most common and natural pattern, you e.g. need to take an explicit care of
the upgrade path (Obsoletes etc.).  But overall looks like OK to me as well (IMVHO a matter of taste).

Tiboris:
@praiskup does it make sense to have mrack-core as a dependency of mrack if it is already transitive of each dependent package?

praiskup:
>  does it make sense to have mrack-core as a dependency of mrack

@Tiboris technically it is not necessarily, when the main (mrack or mrack-all) package is only supposed to be a "metapackage".   If not a metapage (== contains at least some files that depend on core) it is better to stay explicit (even though the dependency could already exist implicitly, generated by rpmbuild itself ... I'd have to test).

praiskup:
> So what happened to the mrack-all and python3- prefix?

+1 for python3- prefixes

Tiboris:
@pvoborni I have fixed the python prefix and updated spec file to reflect things we agreed on in our discussion, shortly:

Splitting mrack to:
- mrack  # all the supported providers and cli
- mrack-cli # mrack command with static provider
- python3-mracklib # core mrack libraries with static provider
- python3-mrack-aws 
- python3-mrack-beaker
- python3-mrack-openstack
- python3-mrack-podman
- python3-mrack-virt

@praiskup I hope that the python prefix is fine now I did not find any directive for such prefix thus i hard-coded it did i tried poorly (to find the directive)?

Tiboris:
Also I have updated `README.md` to suggest installation steps when only a subset of functionality is required.

Tiboris:
PR is ready for review.

mrack.spec
```diff
@@ -1,59 +1,141 @@
...
+
+%{?python_provide:%python_provide %{name}}
+
+Requires:       %{name}-cli == %{version}-%{release}
+Requires:       python3-%{name}lib == %{version}-%{release}
+Requires:       python3-%{name}-aws == %{version}-%{release}
+Requires:       python3-%{name}-beaker == %{version}-%{release}
+Requires:       python3-%{name}-openstack == %{version}-%{release}
+Requires:       python3-%{name}-podman == %{version}-%{release}
+Requires:       python3-%{name}-virt == %{version}-%{release}
```
praiskup, line 33: Requires need to be specified before the %description tag (otherwise they are just part of %description).

Tiboris: Fixed

praiskup:
> @praiskup I hope that the python prefix is fine now I did not find any directive for such prefix thus i hard-coded it did i tried poorly (to find the directive)?

No, seems fine -  python3- prefix is usually hard-coded in every %package, %description, %files tag....

README.md
```diff
@@ -121,9 +121,36 @@ pip install mrack
...
+sudo dnf install mrack-cli python3-mrack-aws
+
+# install mrack library to be later used as an import with support of OpenStack provider:
+sudo dnf install python3-mrack-openstack
+
 ```
+---
+**NOTE**
+
+Above command will install `mrack` package which depends (contains) following packages:
```
pvoborni, line 140: `depends` -> `depends on` ?

pvoborni (approved):
The changes looks good to me - all the names and deps. 

But I did not test it, thus might have missed something.

README.md
Tiboris: Let me fix this

mrack.spec
```diff
@@ -1,59 +1,145 @@
...
 %license LICENSE
 %doc README.md
-%{_bindir}/mrack
-%{python3_sitelib}/%{srcname}
-%{python3_sitelib}/%{srcname}-%{version}-py%{python3_version}.egg-info
+%{python3_sitelib}/%{name}
+%{python3_sitelib}/%{name}-%{version}-py%{python3_version}.egg-info
+%exclude %{python3_sitelib}/%{name}/providers/utils/{,__pycache__/}osapi.py
+%exclude %{python3_sitelib}/%{name}/providers/{,__pycache__/}{%{provider_plugins}}.*
+%exclude %{python3_sitelib}/%{name}/transformers/{,__pycache__/}{%{provider_plugins}}.*
```
praiskup, line 121: I'd personally do `%dir %{python3_sitelib}/%{name}` and then drop the %exclude statements

mrack.spec
```diff
@@ -1,59 +1,145 @@
...
+
+Requires:       %{name}-cli == %{version}-%{release}
+Requires:       python3-%{name}lib == %{version}-%{release}
+Requires:       python3-%{name}-aws == %{version}-%{release}
+Requires:       python3-%{name}-beaker == %{version}-%{release}
+Requires:       python3-%{name}-openstack == %{version}-%{release}
+Requires:       python3-%{name}-podman == %{version}-%{release}
+Requires:       python3-%{name}-virt == %{version}-%{release}
+
+%global __requires_exclude asyncopenstackclient
```
praiskup, line 28: Please document why you are doing this :-)

mrack.spec
Tiboris: As discussed this would create a long list of includes doing it this way... Lets keep the excludes only ?

mrack.spec
Tiboris: Updated

mrack.spec
praiskup: taking back this one, the list of "include" statements would be too large...

praiskup:
looks fine to me

PR #191, opened 2022-09-30
feat: Support package splitting in the mrack

Tiboris:
Change a code bit to support future mrack division to multiple packages


dav-pascual (approved):
Changes look good from my POV (didn't test)
Thanks @Tiboris

PR #195, opened 2022-10-11
fix(Beaker): connection to hub timing out

Tiboris:
mrack was failing with an exception when
Beaker hub was not replying to requests for
some time. With this we accept the fact that
hub may not respond and we use previous
result and continue the wait for resource.

### DIFF
--- a/src/mrack/providers/beaker.py
+++ b/src/mrack/providers/beaker.py
@@ -293,27 +293,39 @@ class BeakerProvider(Provider):
         bkr_res = {}
         prev_status = ""
         job_url = ""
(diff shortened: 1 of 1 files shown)


XiyangDong (commented):
Looks good to me, shall we quickly test that the change does not break beaker provisioning?

Tiboris:
Let me re-base and run something.

XiyangDong:
I am ok with merging this if @dav-pascual is also in favor of that.

dav-pascual (approved):

dav-pascual:
Fine by me as well!

PR #196, opened 2022-10-14
fix(Beaker): traceback when hub is not accessible at session creation

Tiboris:


kaleemsiddiqu (approved):
local run is fine, so adding the ack

PR #200, opened 2022-10-20
feat(pytest-multihost): Allow extended host config to be passed from …

jakub-vavra-cz:
…metadata.

Extended host configuration under "config" is needed for framework build upon pytest multihost https://github.com/pbrezina/sssd-tests-poc

### DIFF
--- a/src/mrack/outputs/pytest_multihost.py
+++ b/src/mrack/outputs/pytest_multihost.py
@@ -67,6 +67,7 @@ class PytestMultihostOutput:
             "username",
             "password",
             "host_type",
+            "config",
         ]
 
         # Use values from provisioning-config mhcfg section - as default values


pvoborni:
This would be a correct patch if the config was a general pytest-multihost attribute usable by multiple projects. But atm it is just  a project-specific attribute for sssd tests POC. Adding tested-project-specific extensions is something we should not do as every new use-case would like something else.   

I'm not saying we cannot merge this eventually, but right now it seems pre-mature.

For project specific attributes mrack supports the following way, e.g. :

```yaml
domains:
  - name: example.test
    type: ipa
    hosts:
      - name: server.example.test
        role: ipa
        group: ipaserver
        os: rhel-8.6
        pytest_multihost: # section for custom pytest multihost attributes
            no_ca: yes
            something_else: for_fun
```

Thus adding a `config` dict into the `pytest_multihost` dict will probably do the trick.

jakub-vavra-cz:
@pvoborni , thanks for clarification, I was just ytrying tre remedy vanishing piece of config.
Will the content of pytest_multihost end directly in the host or will there be a pytest_multihost in the resulting mhc.yaml?

jakub-vavra-cz:
Looks like this change is not needed.

PR #201, opened 2022-10-20
feat(AWS): Add multiple subnet support & IPs availability check

dav-pascual:
- Now mrack expects either subnets_id (prioritized) or subnet_id field in the provining config file in AWS key field. A list of subnets can be specified where host can be provisioned. It is expected that the subnets are in the same VPC.
- mrack checks that the amount IPs available with all the subnets specified are enough to provision all the host concurrently. Otherwise provisioning will retried after a wait and/or cancelled.
- Add rule to pylintrc file to allow more local variables in methods.

Signed-off-by: David Pascual <davherna@redhat.com>


src/mrack/providers/aws.py
```diff
@@ -209,9 +209,41 @@ async def validate_hosts(self, reqs):
...
+        except ClientError:
+            err_msg = (
+                f"{log_msg_start} Error retrieving " f"info from subnet: {subnet_id}"
+            )
+            logger.error(err_msg)
+            raise
+
+        return subnet.available_ip_address_count
+
     async def can_provision(self, hosts):
```
dav-pascual, line 225: One observation about this method:
It is not part of this PR, but it is possible also to check CPU limits (like we do on Openstack).
e.g. vCpus limit number rigth now is 640.
Ref: https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-on-demand-instances.html#ec2-on-demand-instances-limits  

This is not limiting rigth now, but just in case, it could be useful in the future.

dav-pascual:
Another note: 
I went with the most flexible solution, that is, The amount of IPs available for provisioning will be the sum of the available IPs in the provided subnets.

src/mrack/providers/aws.py
```diff
@@ -209,9 +209,47 @@ async def validate_hosts(self, reqs):
...
+            for subnet_id in subnet_ids:
+                available_ips += self.get_subnet_available_ips(subnet_id, log_msg_start)
+        elif subnet_id := hosts[0].get("subnet_id"):
+            available_ips += self.get_subnet_available_ips(subnet_id, log_msg_start)
+        else:
+            logger.warning(
+                f"{log_msg_start} No subnet/s specified. "
+                "Could not check IP addresses availability."
+            )
+            return True
```
dav-pascual, line 245: Default behaviour when no subnet is specified is ignore it and go forward (same as before).
What are your thougths on this?

Tiboris: If we do not have such information in the provisioning config this behavior of continue as is is a good move IMO maybe it does not even have to be warning but just info.

There are few concerns I have about the specific details. Lets meet I will present them to you.

I think we should do similar logic later as in OpenStack which open us the door to the network picking methods or functions. We should have the information about the networks in the provisioning config right?

So in general, you should extract "from config hierarchy" the subnet_id/subnet_ids and then just use the subnet or if multiple specified pick one (in case there is list in provisioning config and host do not specify the network).
However this part should be in different method bound to the create_server()

src/mrack/providers/aws.py
```diff
@@ -209,9 +209,47 @@ async def validate_hosts(self, reqs):
...
+
+        logger.info(
+            f"{log_msg_start} There are "
+            f"{available_ips} IPs available and {len(hosts)} hosts to be provisioned"
```
Tiboris, line 249: Personal note: Tiny detail, I would prefer printing out availability per network in cycle.

src/mrack/transformers/aws.py
```diff
@@ -72,6 +72,9 @@ def create_host_requirement(self, host):
             "spot": self._find_value(host, "spot", None, None),
             "delete_volume_on_termination": del_vol,
         }
-        if self.config.get("subnet_id"):
+        if self.config.get("subnet_ids"):
+            req["subnet_ids"] = self.config.get("subnet_ids")
+        elif self.config.get("subnet_id"):
             req["subnet_id"] = self.config.get("subnet_id")
```
Tiboris, line 78: What do you think about this? 
```suggestion
        req["subnet_ids"] = self.config.get("subnet_ids", [])
        req["subnet_id"] = self.config.get("subnet_id", "")
```
With get and default you do not need if/elif statements and the empty default would help later to check if the value has been really set.

Tiboris (changes requested):
Hello @dav-pascual Thanks a lot for the patch.
The code has decent quality but i left some logic enhancements or suggestions which we can talk about. Please ping me if any of comments is not clear We can meet and discuss the details.

src/mrack/providers/aws.py
```diff
@@ -265,8 +303,21 @@ async def create_server(self, req):
             ],
             "TagSpecifications": [{"ResourceType": "instance", "Tags": taglist}],
         }
-        if specs.get("subnet_id"):
-            request["SubnetId"] = specs.get("subnet_id")
+
+        if subnet_ids := specs.get("subnet_ids"):
+            for subnet_id in subnet_ids:
+                available_ips = self.get_subnet_available_ips(subnet_id, log_msg_start)
```
pvoborni, line 309: Atm the biggest problem of AWS provider (i.e. not this PR) is that it is not truly asynchronous as e.g. OpenStack provider. It would be good to get to the async nature (but I'm not pushing for it here). E.g. in the method which should be asynchronous we are calling 2 blocking synchronous calls - getting the IPs and and then provisioning the instance.

I'd rather skip the getting of IPs here. It will probably not be reliable anyway, thus should be OK to do once at the beginning of provisioning, e.g. in prepare_provisioning then use it in can_provision. We could use similar methods in distribution as in openstack. But I think it is good enough to get all IPs from a single network and fail if there isn't any available.  The limit can be re-checked/updated on failure of provisioning.

pvoborni (commented):
I've commented with my view, but I don't insist on getting there in this PR - it's more for providing a different perspective.

.pylintrc
```diff
@@ -16,3 +16,6 @@ disable=fixme,too-many-instance-attributes,no-self-use,attribute-defined-outside
 
 # Minimum lines number of a similarity.
 min-similarity-lines=11
+
+# Maximum number of locals for function / method body
+max-locals=25
```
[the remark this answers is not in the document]
dav-pascual: create_server from providers/aws.py

I see that we use `# pylint: disable-msg=too-many-locals ` in some other functions, that's why I think it's a good idea to raise max-local value from 15 (default) to 25.

src/mrack/providers/aws.py
dav-pascual: My idea was having a pool of IPs from the specified subnets.
Example:
- We want to provision 100 host.
- SubnetA has 80 addresses and SubnetB has 50 addresses.
- We start provisioning in SubnetA. Once the are no addresses left in SubnetA, we continue provisioning the remaining 20 hosts in SubnetB. This way we have more capacity than picking a subnet at the beginning of the provisioning. It'd OK to print per Network as additional info, but what matter in this implementation is the global count.

src/mrack/providers/aws.py
dav-pascual: I'm OK with this async approach, but I think it would incompatible with https://github.com/neoave/mrack/pull/201#discussion_r1005968388 
So we would have to pick.

.pylintrc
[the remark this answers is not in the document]
Tiboris: I am fine with it

src/mrack/providers/aws.py
Tiboris: If you create something like openstack has:
`python
self.ip_availabilities = {
     "subnet_id_1" = 20,
     "subnet_id_2" = 80,
}
 here in can provision 
you can later reuse it and do not query the pools again that was the point i wanted to point out in our meeting and at the end of this function maybe not in cycle as i initially said you could print nicely this dictionary in similar style we have in openstack already.

src/mrack/providers/aws.py
Tiboris: I left a comment there for you @dav-pascual lets continue to work that both are possible :muscle:

src/mrack/providers/aws.py
dav-pascual: Comments addressed.
We can now specify subnet/s in provider config (metadata) as well as in provisioning config.
The reason for keeping "subnet_id" is to support old metadata files, but it'll be legacy and "subnet_ids" should be used.

The priority order is: 
subnet_ids in provider config > subnet_id in provider config > subnet_ids in provisioning config > subnet_id in provisioning config

src/mrack/providers/aws.py
```diff
@@ -209,9 +209,47 @@ async def validate_hosts(self, reqs):
         """Validate that host requirements are well specified."""
         return
 
+    def get_subnet_available_ips(self, subnet_id, log_msg_start):
```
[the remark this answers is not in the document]
dav-pascual: IMO I would keep it for more customization

src/mrack/providers/aws.py
```diff
@@ -209,9 +209,47 @@ async def validate_hosts(self, reqs):
...
+        """Get number of IPs available in a subnet."""
+        try:
+            subnet = self.ec2.Subnet(subnet_id)
+        except ClientError:
+            err_msg = (
+                f"{log_msg_start} Error retrieving " f"info from subnet: {subnet_id}"
+            )
+            logger.error(err_msg)
+            raise
```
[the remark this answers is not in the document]
dav-pascual: I went with logger.warning suggestion.
Changes added

src/mrack/providers/aws.py
dav-pascual: Comments addressed!
I used something similar to your suggestion

src/mrack/transformers/aws.py
dav-pascual: Logic extracted to external function

src/mrack/providers/aws.py
dav-pascual: I pushed new changes addressing the comments.
I'm also taking a look at the async way.

src/mrack/providers/aws.py
dav-pascual: @Tiboris sorry, I accidentally deleted your comment in this thread

Tiboris: No worries

src/mrack/providers/aws.py
dav-pascual: I deleted the sync calls to AWS API from create_server. 
I also made the loop calling to AWS API in `can_provision` async with gather function.
Please let me know if anything can be improved further to make it async.

I have observed that we could make the loop in `_provision_base` func async similarly as I have done, although that would be better for another PR.

src/mrack/providers/aws.py
```diff
@@ -209,9 +212,87 @@ async def validate_hosts(self, reqs):
...
+                    f"{log_msg_start} No subnet/s specified for host {host['name']}."
+                )
+
+        # Get available IPs from AWS for every subnet
+        ips_count_wait = [
+            self.get_subnet_available_ips(subnet_id, log_msg_start)
+            for subnet_id in ip_availabilities
+        ]
+        ips_count_result = await asyncio.gather(*ips_count_wait)
+        if not all(ips_count_result) > 0:
```
Tiboris, line 268: This does not seem like valid logic operation all returns True/False i believe. so enough is to ask for all()
```suggestion
        if not all(ips_count_result):
```

Tiboris (changes requested):
@dav-pascual thanks a lot for the updates. I did comment some things which i found reasonable.
Please have a look and provide me with your opinion.

src/mrack/providers/aws.py
```diff
@@ -209,9 +212,87 @@ async def validate_hosts(self, reqs):
...
+                    ip_availabilities[subnet_id] -= 1
+                    no_ips_flag = False
+                    break
+            if no_ips_flag:
+                logger.info(
+                    f"{log_msg_start} Not enougn IP addresses available "
+                    f"in subnet/s: {subnet_ids}"
+                )
+                return False
+            no_ips_flag = True
```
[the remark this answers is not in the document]
dav-pascual: With this implementation, it's supposed to check every subnet provided (in subnet_ids) before deciding that it can't be provisioned, that's why I wanted to break to the outer loop.
I have refactored the code using `For-Else` [syntax](https://book.pythontips.com/en/latest/for_-_else.html#for-else) to get rid of the variable.

src/mrack/providers/aws.py
```diff
@@ -209,9 +212,87 @@ async def validate_hosts(self, reqs):
...
+                )
+
+        # Get available IPs from AWS for every subnet
+        ips_count_wait = [
+            self.get_subnet_available_ips(subnet_id, log_msg_start)
+            for subnet_id in ip_availabilities
+        ]
+        ips_count_result = await asyncio.gather(*ips_count_wait)
+        if not all(ips_count_result) > 0:
+            return False
```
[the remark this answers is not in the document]
dav-pascual: You're rigth, this should be checked in the single_subnet_prov and mult_subnets_prov loops instead.
I have refactored the code to print the error and do the rigth checks.

src/mrack/providers/aws.py
dav-pascual: Fixed

src/mrack/providers/aws.py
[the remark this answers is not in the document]
Tiboris: Thanks

src/mrack/providers/aws.py
Tiboris: Fine for now I would say

src/mrack/providers/aws.py
```diff
@@ -209,8 +212,87 @@ async def validate_hosts(self, reqs):
...
+        except ClientError:
+            logger.warning(
+                f"{log_msg_start} Error retrieving info from subnet: {subnet_id}"
+            )
+            return 0
+
+        logger.debug(
+            f"{log_msg_start} Subnet Id: {subnet_id}\n"
+            f"\tHas {subnet.available_ip_address_count} IP addresses available"
+        )
```
Tiboris, line 228: This is a nitpick but It would be nice to cahgne output to something like (without even `log_msg_start`):
```suggestion
        logger.debug(f"Subnet Id: {subnet_id}")
        logger.debug(f"  available: {subnet.available_ip_address_count}")
```

To look similar like:
```
   2022-10-27 01:35:04,175 mrack.providers.openstack DEBUG Network: foo_network
    2022-10-27 01:35:04,175 mrack.providers.openstack DEBUG   total: 251
    2022-10-27 01:35:04,175 mrack.providers.openstack DEBUG   used: 105
    2022-10-27 01:35:04,175 mrack.providers.openstack DEBUG   available: 146
```

src/mrack/providers/aws.py
```diff
@@ -209,8 +212,87 @@ async def validate_hosts(self, reqs):
...
+                    mult_subnets_prov.append(subnet_ids)
+                    for subnet_id in subnet_ids:
+                        if subnet_id not in ip_availabilities:
+                            ip_availabilities[subnet_id] = 0
+            else:
+                logger.debug(
+                    f"{log_msg_start} No subnet/s specified for host {host['name']}."
+                )
+
+        # Get available IPs from AWS for every subnet
```
Tiboris, line 262: maybe in addition to above comment:
```suggestion
        # Get available IPs from AWS for every subnet
        logger.info(f"{log_msg_start} Checking IP availability per available network")
```

Tiboris (changes requested):
Thansk @dav-pascual for the patch, logic changes are looking better now.
This is hopefully last comment and consider this a nitpick which is nice to have.
based on log:
```
2022-10-31T22:07:33     2022-10-31 22:05:25,425 mrack.providers.provider INFO AWS Checking available resources
2022-10-31T22:07:33     2022-10-31 22:05:25,614 mrack.providers.aws DEBUG AWS Subnet Id: subnet-08bd210025a255be9
2022-10-31T22:07:33             Has 56 IP addresses available
2022-10-31T22:07:33     2022-10-31 22:05:25,706 mrack.providers.aws DEBUG AWS Subnet Id: subnet-0119dfca46c7527f4
2022-10-31T22:07:33             Has 247 IP addresses available
2022-10-31T22:07:33     2022-10-31 22:05:25,706 mrack.providers.provider INFO AWS Resource availability: OK
```
which has weird log format

src/mrack/providers/aws.py
dav-pascual: Added

src/mrack/providers/aws.py
dav-pascual: I have update it to look like this (I left log_msg_start bc I think it's useful to spot things in the log):
```
    2022-11-02 11:37:21,077 mrack.providers.aws DEBUG AWS Subnet subnet-08bd210025a255be9
    2022-11-02 11:37:21,483 mrack.providers.aws DEBUG AWS   available: 58
    2022-11-02 11:37:21,484 mrack.providers.aws DEBUG AWS Subnet subnet-0119dfca46c7527f4
    2022-11-02 11:37:21,720 mrack.providers.aws DEBUG AWS   available: 251
```

Tiboris:
Thanks for the patience all the changes and the feature. lets merge this @dav-pascual !

dav-pascual:
Thanks for your feedback and review!! @Tiboris :)

Issue #186, opened 2022-09-19
 looks OK. Thus approving.


Tiboris:
I've tested this locally, it seems to work. The code also looks OK. Thus approving.

That said, it would be super to have unit tests also for the other parts touched in this MR, namely:  `task`, `retention_tag`, `product`  a ssh part of `ks_append`. But these can be added separately.

_Originally posted by @pvoborni in https://github.com/neoave/mrack/pull/185#pullrequestreview-1112419443_

PR #206, opened 2022-11-02
test: extend beaker tests

b3lix:
test(Beaker): add check for additional keys
    
    Add check for additional keys [tasks, redention_tag, product]
    to beaker transformer tests.
    Resolve: https://github.com/neoave/mrack/issues/186

refactor(Beaker): transformer test to use dictionary
    
    Refactor beaker transformer test to use dictionary instead of multiple
    parameters in function call. This increase readability of test.

Resolve: https://github.com/neoave/mrack/issues/186

### DIFF
--- a/tests/unit/test_beaker_transformer.py
+++ b/tests/unit/test_beaker_transformer.py
@@ -114,67 +114,80 @@ class TestBeakerTransformer:
 
     @pytest.mark.asyncio
     @pytest.mark.parametrize(
-        "meta_host,exp_distro,exp_variant,exp_ks_meta,exp_ks_append,exp_whiteboard,exp_prio",  # noqa: E501
+        "meta_host, exp_values",  # noqa: E501
         [
             (
                 fedora,
-                "Fedora-36%",
-                "Server",
-                "FEDORA_HOST_KS_META",
-                default_ks_append,
-                default_whiteboard,
-                default_prio,
+                {
+                    "distro": "Fedora-36%",
+                    "variant": "Server",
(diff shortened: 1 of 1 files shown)


tests/unit/test_beaker_transformer.py
```diff
@@ -114,67 +117,97 @@ async def create_transformer(self, legacy=False):
...
+                    "variant": "BaseOS",
+                    "ks_meta": "PROV_CONF_RHEL86_KS_META",
+                    "ks_append": [
+                        "%post\ncat /etc/redhat-release\nwget redhat.com\n%end"
+                    ],
+                    "whiteboard": default_whiteboard,
+                    "priority": default_prio,
+                    "tasks": default_tasks,
+                    "retention_tag": default_retention_tag,
+                    "product": default_product,
```
Tiboris, line 181: Do you think you can add some other than default values to one or two hosts ? 
I mean:
```
retention_tag
product
```
there I can see only defaults.

Tiboris (changes requested):
Hi @b3lix !

Thanks for the patch it is very nice to see new contributor.
Also adding tests is much appreciated. Please find inline comment I would like you to try and update slightly the test.

tests/unit/test_beaker_transformer.py
Tiboris: Thanks for the additions. Looks fine now.

b3lix:
Thank you for your help @Tiboris. Glad to contribute!

PR #208, opened 2022-11-03
chore: disable automatic runtime deps discovery for rpm build

pvoborni:
As it is adding all dependencies to all rpms and ignoring the individual nuaces between packages. I.e. without this, we cannot install python3-mracklib without the rest.


pvoborni:
I think we can avoid using this if we split the package in setup.py by using optional dependencies. https://setuptools.pypa.io/en/latest/userguide/dependency_management.html#optional-dependencies

Tiboris (approved):
Building and dependencies works for me. I have tried build and installations.

PR #209, opened 2022-11-03
chore: rpm and clean-rpms targets in Makefile

pvoborni:
To make buiding of rpms easier - for adhoc testing/development and sort of work as a documentation how to do it as it is easily forgotten when not doing often.


Tiboris (approved):
Thanks a lot this is helpful.

PR #212, opened 2022-11-09
fix: Do not use deprecated asyncio.coroutine wrapper

Tiboris:
Python 3.11.0 completely dropped support for the
asyncio.coroutine. As it was already deprecated
we should use the proper way of decorating the
click functions in the mrack run.py.


pvoborni (approved):
LGTM, I've tested the all the basic commands and it did run well.

PR #214, opened 2022-11-15
chore: Use personal GH token to release mrack

Tiboris:
When branch protection is enabled only real
user (not gh-actions as it is bot) can push
to the main branch when there is Pull Request
required and approvals for particular PR.
With exception added to the users which can bypass the protection rules we should be able to release
with personal GH token generated for the user.


pvoborni (approved):
After reading through https://github.com/semantic-release/github/issues/175 , this change looks good to me (though I don't know if it indeed works) :)

Tiboris:
Yup i need to test it somehow

PR #215, opened 2022-11-16
fix: validate ownership and lifetime only for AWS and OpenStack

pvoborni:
Limitin only for certain providers is needed as the owner doesn't matter for static (vm already present somewhere, could be owned by completely different person), podman, libvirst (vm running on current machine, no need for owners).

But it makes sense for any cloud like OpenStack, AWS, Beaker.

Enabled only for AWS, OpenStack as Beaker doesn't work with ownership yet.

### DIFF
--- a/src/mrack/transformers/aws.py
+++ b/src/mrack/transformers/aws.py
@@ -90,6 +90,11 @@ class AWSTransformer(Transformer):
 
         return []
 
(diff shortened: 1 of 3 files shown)


kaleemsiddiqu:
tested this patch and its working.

kaleemsiddiqu (approved):

PR #216, opened 2022-11-21
fix: raise Validation error when validation fails and use MrackError

Tiboris:
This is bit cosmetic change, adding MrackError class which is parent of all errors so later we just catch the MrackError exceptions.


dav-pascual (approved):
This changes for error handling LGTM! (haven't tested) 
Thanks @Tiboris

Tiboris:
I have run 2 regression runs in our CI should be okay to merge, thanks @dav-pascual for looking into the code.

PR #219, opened 2022-11-23
fix: make db file and provisioning file optional

Tiboris:
When initializing the global context these files
should be passed to the init as optional param
when path is specified in mrack cli tool.
Otherwise mrack should only load the data from
the mrack.conf file records and rely on paths
set there for db and provisioning-config.

### DIFF
--- a/src/mrack/context.py
+++ b/src/mrack/context.py
@@ -52,7 +52,7 @@ class GlobalContext:
         """Get ProvisioningConfig object."""
         return self.metadata
 
-    def init(self, mrack_config, provisioning_config, db_file):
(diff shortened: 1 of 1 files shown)


dav-pascual (approved):
LGTM!

pvoborni:
What is the context/purpose of this? Don't know what problem it solves.

pvoborni:
That said the code looks good to me and it seems that the method was written in a way to have this optional.

Tiboris:
When initializing the global context we might want to only use mrack.conf file which without optional params was not possible. It would be really useful when using mrack as lib to have one source of truth and no need to pass 3 options to lib call.

PR #223, opened 2022-11-30
fix: Owner requirement boolean parsing from string

Tiboris:
When parsing boolean values from string using bool() builtin funtion string which is not empty returns True Writing a util function to convert such strings
to proper boolean values. NOTE: distutil implementation: from distutils.util import strtobool
will be deprecated and removed in future python releases thus implementing a funtion ourselves might be
future proof solution at this time.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -9,6 +9,9 @@ multi_line_output = 3
 markers = [
     "etc_hosts: passes path for etc_hosts fixture",
 ]
+pythonpath = [
+  "./src"
+]
 
 [tool.semantic_release]
 version_source = "commit"
--- a/src/mrack/config.py
+++ b/src/mrack/config.py
@@ -22,6 +22,7 @@ import os
 from configparser import ConfigParser, NoOptionError, ParsingError
 
 from mrack.errors import ConfigError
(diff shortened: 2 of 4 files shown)


pvoborni:
`value_to_bool` looks like, that it is a perfect candidate for a unit test ;)

Tiboris:
I thought that when affected mrack util unit tests were not wrapped by Class It caused that they were not run. 
The old run of pytest:

https://dev.azure.com/neoave/mrack/_build/results?buildId=816&view=logs&j=27008f10-b491-547d-a654-20cfee07d03e&t=73113a39-43fd-502e-b4fb-b68667aceda6&l=330

```
======================= 81 passed, 22 warnings in 1.32s ========================
```

Adding the class and waiting for new run.

@miskopo Could you try to help me find what I am missing?

kaleemsiddiqu:
Though not related with this but we should one test for osp env as well like for get_username(host_win_aws, metahost_win, provisioning_config)
Rest LGTM

kaleemsiddiqu (approved):

Tiboris:
Thanks for review !

PR #225, opened 2022-12-07
fix(AWS): refactor sources to be py3.6 compatible and add epels to packit

Tiboris:
chore(Packit): enable epel-8 and epel-9 updates and tests

    fix(AWS): refactor sources to be py3.6 compatible
    
    When considering mrack to be part of
    the tmt which is used even in epel8 and epel9
    we might be limited to not use latest python
    features which one of them is := opearator.

### DIFF
--- a/.packit.yaml
+++ b/.packit.yaml
@@ -27,25 +27,22 @@ actions:
 jobs:
   - job: copr_build
     trigger: pull_request
-    targets: [fedora-all]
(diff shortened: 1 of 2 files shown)


Tiboris:
@dav-pascual Please have a look as I am basically touching your code.

dav-pascual (approved):
Ack! @Tiboris

pvoborni (approved):

PR #227, opened 2022-12-12
fix: Add several c8s fixes for missing deps and decorators

Tiboris:
fix: Add cache decorator for older python versions.
    
    fix(mrack.spec): Missing dependency in c8s for beaker-client

### DIFF
--- a/mrack.spec
+++ b/mrack.spec
@@ -61,7 +61,12 @@ Requires:       python3-botocore
 %package -n     python3-%{name}-beaker
 Summary:        Beaker provider plugin for mrack
 Requires:       python3-%{name}lib = %{version}-%{release}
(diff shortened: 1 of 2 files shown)


Tiboris:
c8s run is now green: https://dashboard.packit.dev/results/testing-farm/153647

Tiboris:
Should I remove temp commits? @pvoborni

pvoborni (changes requested):
the c8s fixes looks good to me. But I think the smoke test has issues in it (will comment in the other PR).

Tiboris:
I will remove the 4 commits from other PR then it should be fine do you agree? and I will rebase #226 on top of this one.

PR #229, opened 2023-01-03
feat: Do not use same sleep for every mrack run

Tiboris:
feat: Do not use same sleep for every mrack run
    
    Adding logic to randomize the timeout to basically
    set the waiting timeout to vary from 45-75 minutes randomly
    and the poll time to 45-75 seconds accordingly
    so if there are multiple parallel runs of mrack they wait
    different amount time before poll and increase chance to get resources
    quicker than other concurrent runs of mrack requests
    
    Add cooldown after error host was found to wait before
    another retry and resource check cycle for random time
    between one poll sleep and 2* poll sleep (45-75 s is one poll)
    
    Add utilization method template to decide whether to
    destroy all machines or to keep them while re-provisioning
    to free some resources when there is provided highly utilized.
    
feat(AWS): Add utilization check method
    
feat(OpenStack): Add utilization check method
    
    minor fix: typo _opentack -> _openstack

refactor(OpenStack): make private openstack methods truly private
    
    renaming the openstack private methods to be private
    openstack had too many public methods and this seems
    to be proper fix, to have public just the needed ones.
    
### DIFF
--- a/src/mrack/providers/provider.py
+++ b/src/mrack/providers/provider.py
@@ -15,6 +15,7 @@
 """General Provider interface."""
 import asyncio
 import logging
+import random
 import socket
 from datetime import datetime, timedelta
 
@@ -40,6 +41,7 @@ class Provider:
         """Initialize provider."""
         self._name = "dummy"
         self.dsp_name = "Dummy"
+        self.timeout = 60
         self.max_retry = 1
         self.strategy = STRATEGY_ABORT
         self.status_map = {"OTHER": STATUS_OTHER}
@@ -55,6 +57,10 @@ class Provider:
 
     async def can_provision(self, hosts):
         """Check that provider has enough resources to provision hosts."""
+        raise NotImplementedError()
+
+    async def utilization(self):
+        """Check percentage utilization of given provider."""
         raise NotImplementedError()
 
     async def create_server(self, req):
@@ -227,17 +233,21 @@ class Provider:
         return success_hosts, error_hosts
 
     async def _provision_base(
-        self, reqs, res_check_timeout=60, res_busy_sleep=10
+        self,
+        reqs,
+        timeout=None,
     ):  # pylint: disable=too-many-locals, too-many-branches, too-many-statements
         """Provision hosts based on list of host requirements.
 
         Main function which does provisioning and validation.
         Parameters:
             reqs - dictionary with requirements for provider
-            res_check_timeout (default 60) - timeout (minutes) to wait for resources
-            res_busy_sleep (default 10) - time to wait before checking again (minutes)
+            timeout - base timeout (minutes) to wait for resources
         """
-        log_msg_start = f"{self.dsp_name}"  # use display name at log message begining
+        if not timeout:
+            timeout = self.timeout
+
+        log_msg_start = f"{self.dsp_name}"  # use display name at log message beginning
         logger.info(f"{log_msg_start} Validating host(s) definitions")
         if not reqs:
             raise ProvisioningError(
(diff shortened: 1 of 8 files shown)


Tiboris:
Oh i forgot to rewrite calls in test code. Will do

Tiboris:
Failure cauised by:
```
Error: 
 Problem: package mrack-1.12.3-4.el8.noarch requires python3-mrack-beaker = 1.12.3-4.el8, but none of the providers can be installed
  - cannot install the best candidate for the job
  - nothing provides beaker-client needed by python3-mrack-beaker-1.12.3-4.el8.noarch
```
Possible resolution is to suggest or recommend the package instead of require to workaround such errors.

Tiboris:
Depends on https://github.com/neoave/mrack/pull/233

dav-pascual:
Hey @Tiboris 
These changes look good at first sight.
Could you help me understand the different sleep times that you've put in place and what they solve?

- `asyncio.sleep(slow_down)` in `_provision_base`: This slows down the provisioning at the beginning using a variable time to sleep so we prevent mrack executions to provision at the same time.
- `await asyncio.sleep(timeout * 10)` in `_provision_base`: Sleep time to wait to check if resources are available again.
- `await asyncio.sleep(cooldown) `in `strategy_retry`: In this case we haven't been able to provision all resources despite the previous waits, so we probably are in a race condition with concurrent runs, and thus we free all resources and try again in `cooldown` minutes.

Please correct me or complement the above :slightly_smiling_face:

src/mrack/providers/provider.py
```diff
@@ -408,23 +435,58 @@ async def strategy_retry(self, reqs):
...
+                    if host.error.get("code", None) == 500:  # 500 = SERVER ERROR
+                        logger.info(
+                            f"{log_msg_start} Provider is out of resources, "
+                            "increasing cooldown time before another retry"
+                        )
+                        out_of_resources = True
+
+                if await self.utilization() >= max_utilization or out_of_resources:
+                    # delete all hosts when there is high utilization of provider
+                    error_hosts += success_hosts
```
dav-pascual, line 472: Just wondering: Wouldn't it be better to check in addition:

if percentage of hosts already provisioned is low, then free resources. If our host list is close to be completely provisioned then we would not free resources.

This way we would get optimized result across concurrent runs?

src/mrack/providers/provider.py
```diff
@@ -408,23 +435,58 @@ async def strategy_retry(self, reqs):
...
                 logger.info(f"{log_msg_start} {err}")
+                out_of_resources = False
                 for host in error_hosts:
                     logger.error(f"{log_msg_start} Error: {str(host.error)}")
+                    if host.error.get("code", None) == 500:  # 500 = SERVER ERROR
+                        logger.info(
+                            f"{log_msg_start} Provider is out of resources, "
+                            "increasing cooldown time before another retry"
+                        )
+                        out_of_resources = True
```
dav-pascual, line 468: Question: Do we want to free all resource even if we only got a 500 error out of large list of provision requests?

Tiboris:
@dav-pascual thanks for review and questions.

> - `await asyncio.sleep(cooldown) `in `strategy_retry`: In this case we haven't been able to provision all resources despite the previous waits, so we probably are in a race condition with concurrent runs, and thus we free all resources and try again in `cooldown` minutes.

I would like to correct this one the actual cooldown now is in seconds, this can be edited if needed and if it makes sense.

dav-pascual:
> @dav-pascual thanks for review and questions.
> 
> > * `await asyncio.sleep(cooldown) `in `strategy_retry`: In this case we haven't been able to provision all resources despite the previous waits, so we probably are in a race condition with concurrent runs, and thus we free all resources and try again in `cooldown` minutes.
> 
> I would like to correct this one the actual cooldown now is in seconds, this can be edited if needed and if it makes sense.
> 
> As an answer to other questions:
> 
> With concurrent runs its hard to say how many hosts or what will be the percentage of provisioned i am afraid its very non-exact here removing all is destructive, I agree, however helps jobs in case all have 1 hosts provisioned only (also speculative statement with small change of reproducing).
> 
> I will use your very nicely formulated assumtions which are basicaly 99% correct and add comments to the code with it.
> 
> Again thanks!

@Tiboris Thanks for adding the comments :slightly_smiling_face: 

Your answer makes sense and the rest looks good so I will approve the PR

dav-pascual (approved):

Tiboris:
Thanks for the review @dav-pascual

PR #230, opened 2023-01-17
feat(OpenStack): Provide a way to disable network spreading

Tiboris:
The network spreading feature has aimed to utilize networks usage however for some particular products it might be not optimal to have resources on different networks. This patch is adding options to mrack.conf which can disable the feature and also set the acceptable threshold of network utilization for OpenStack.

network-spread = no/allow/force

### DIFF
--- a/tests/unit/test_openstack.py
+++ b/tests/unit/test_openstack.py
@@ -11,13 +11,15 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+import os
 from copy import deepcopy
 from unittest import mock
 from unittest.mock import Mock, patch
 
 import pytest
 
-from mrack.errors import MrackError, ValidationError
+from mrack.context import global_context
+from mrack.errors import MrackError, ProviderNotExists, ValidationError
 from mrack.providers.openstack import OpenStackProvider
 
 from .utils import get_data  # FIXME do not use relative import
@@ -75,8 +77,8 @@ net_3 = networks[2]
 net_3["used_ips"] = 989
 net_3["subnet_ip_availability"][0]["used_ips"] = 989
 
-no_usable_traceback_2ips_left = deepcopy(mocked_nets)
-networks = no_usable_traceback_2ips_left["network_ip_availabilities"]
+net_2ips_left = deepcopy(mocked_nets)
+networks = net_2ips_left["network_ip_availabilities"]
 net_1 = networks[0]
 net_1["used_ips"] = net_1["total_ips"] - 1
 net_1["subnet_ip_availability"][0]["used_ips"] = net_1["total_ips"] - 1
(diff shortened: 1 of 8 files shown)


pvoborni:
About the interface of this feature (the config options):

I like it being configurable, the idea with different max utilization is also good. And I also like the concept of a small job which goes into the "single-network mode". 

But I'm thinking if the `force-network-spread` is the best name. I can imagine following scenarios:

1. the job doesn't care about the the used way
2. the job requires spreading
3. the job requires one network - e.g. because  it will fail otherwise

The usecase which we are solving here is #3 but the configuration options doesn't allow it. Spreading will be used if the job is not considered small - so it might be unreliable.  I think that it might be better to fail the job at provisioning, saying there is no suitable network rather then randomly failing in the job as it got a bad network.

For that reason, I'd propose to change the `force-network-spread` into:

- name: `network-spread`
- value: enum (string), `[allow, force, no]` with `allow` being default.

The logic can be:
- if job defines `network-spread: no` then it will always try to find the suitable single network
- if `allow` then it will use single network or spread based on its size and configured treshold
- if `force` then spreading will be used

WDYT?

Tiboris:
This looks like minor change to existing code and sane workflow procedure for provisioning let me re-write a bit to support this.

pvoborni:
Could you also add some tests which demonstrates that it works?  It's easier to test in mock situation than in prod, where you cannot easily simulate the conditions.

Tiboris:
@pvoborni I have added some tests

tests/unit/test_openstack.py
```diff
@@ -180,7 +206,164 @@
...
         net_pools,
         no_usable_traceback_no_ips_left,
-        ValidationError("Error: no available networks for 3 hosts with IPv4"),
+        ValidationError("no available networks"),
+        "mrack.conf",
+        "allow",
+        75,
+    ),
+    (
+        "test3-hosts_force_spread_no_usable_traceback_no_ips_left",
```
pvoborni, line 215: I'm not sure if I'm able to do the review of this. The way the input data is written is very difficult to understand what it is actually testing and if it behaves correctly. I have no idea what are the avail abilities in each of the test cases and it is difficult to calculate in head.

I'd expect a comment for each test saying what it is actually testing. The test name tries to encode it, but it its difficult to decode. An actual sentence would be helpful. 

You also don't need to add all of the cases into single test method. It's possible to create multiple tests which are using some common logic implemented in different helper functions and thus save the number of input args and make the test usable. 

E.g. when testing spreading and no spreading, the most important attributes are number of host and utilization of the network - these are the values to be displayed in the input params.

Generate the network availability structure with a helper method. The input can be list of dicts/tripples with `network_name`,  `total_ips`, `used_ips`.

Tiboris: now the tests should look better

Tiboris: thanks for the commit cherrypicked and now i know what you meant.

tests/unit/test_openstack.py
```diff
@@ -293,37 +238,337 @@ async def test_provision(self):
...
+                net_availability("net2", 100, 50, 4),
+            ],
+            "expected": [
+                "net1 OR net2",
+                "net1 OR net2",
+                "net1 OR net2",
+            ],
+        },
+        {
+            "name": "can_not_satisfy_over_threshold",
```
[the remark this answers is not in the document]
Tiboris: Hmm as we were discussing this on our slack, this is disabling spread option and all neworks here are above acceptable threshold. So i think in this case this error is fine.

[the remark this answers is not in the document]
Tiboris: Based on followup discussion changed behavior and testcases

Tiboris:
Thanks for the review @pvoborni !

PR #231, opened 2023-01-26
test: fix installation of right package in smoke tests

pvoborni:
The smoke test is uninstalling all packages and then trying to install them again to check if dependencies work correctly.

The issue is that version of package created by packit on pull request could be lower from dnf PoV than the one available e.g. in EPEL thus the EPEL version is used istead of the one which we are testing.

Thus c8s test is failing as it is picking up a package which has incorrect dependencies.

This PR is recording the initially installed package version to use it again on reinstallation and thus avoid the issue.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -61,9 +61,9 @@ repos:
         args: ["--ignore-missing-imports"]
         additional_dependencies: ["types-PyYAML", "types-python-dateutil"]
 
-  - repo: https://github.com/pre-commit/mirrors-pylint
-    rev: v3.0.0a5
-    hooks:
-      - id: pylint
-        additional_dependencies: ["isort[pyproject]"]
-        exclude: ^(docs/|tests/).*$
+  # - repo: https://github.com/pre-commit/mirrors-pylint
+  #   rev: v3.0.0a5
+  #   hooks:
+  #     - id: pylint
+  #       additional_dependencies: ["isort[pyproject]"]
+  #       exclude: ^(docs/|tests/).*$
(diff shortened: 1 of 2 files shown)


.pre-commit-config.yaml
```diff
@@ -61,9 +61,9 @@ repos:
         args: ["--ignore-missing-imports"]
         additional_dependencies: ["types-PyYAML", "types-python-dateutil"]
 
-  - repo: https://github.com/pre-commit/mirrors-pylint
-    rev: v3.0.0a5
-    hooks:
-      - id: pylint
-        additional_dependencies: ["isort[pyproject]"]
-        exclude: ^(docs/|tests/).*$
+  # - repo: https://github.com/pre-commit/mirrors-pylint
```
Tiboris, line 64: Lets create issue from this

Tiboris (approved):
Thanks for the fix and workaround. I have created issue from the pylint.

pvoborni:
I've updated it yet-again as `dev.{PACKIT_PROJECT_BRANCH}` was still lower than e.g. someting which started with `1`. 

```
$ rpmdev-vercmp 1.12.3-dev.pr231 1.12.3-1.20230126174817469017.pr231.4.g887f46d
1.12.3-dev.pr231 < 1.12.3-1.20230126174817469017.pr231.4.g887f46d

$ rpmdev-vercmp 1.12.3-99.dev.pr231 1.12.3-1.20230126174817469017.pr231.4.g887f46d
1.12.3-99.dev.pr231 > 1.12.3-1.20230126174817469017.pr231.4.g887f46d
```

pvoborni:
In general, I think that we have 2 ways to solve the issue
- either have high release number, e.g. the 99 as proposed here. The Q is whether it should be in spec or packit config. I chose packit config as I was not sure how koji update works with it and thus 'pull_request' in packit config should not affect it
- alternative solution would be to have higher version in git than the last release. I.e. bump it after release, but this may not work well with semantic releases functionality (assuming here). Thus the former was chosen.

PR #233, opened 2023-02-01
refactor(OpenStack): refactor openstack method names

Tiboris:
this is only renaming the method names and extracted subset of commits from https://github.com/neoave/mrack/pull/229


dav-pascual:
LGTM

dav-pascual (approved):

PR #234, opened 2023-02-01
feat: Allow domain and global section records for ansible inventory 

Tiboris:
feat(AnsibleInventory): Allow additional domain level ansible inventory values
    
    Add support for domain level ansible_inventory records which
    can be eventually overwritten by host defined values.
    Added unit tests to test this feature.   

    feat(AnsibleInventory): Allow additional global level values
    
    Add support for global level ansible_inventory records
    which can be eventually overwritten by domain and host defined values.
    Minor fix of logig in pytest multihost code - adding deepcopy.
    
    test(AnsibleInventory): global level output values override
    
    Add test for the feature to support global level
    ansible_inventory records which can be eventually
    overwritten by domain and host defined values.

    
### DIFF
--- a/src/mrack/outputs/pytest_multihost.py
+++ b/src/mrack/outputs/pytest_multihost.py
@@ -125,13 +125,13 @@ class PytestMultihostOutput:
                     mhcfg["ad_sub_hostname"] = host["name"]
                     mhcfg["ad_sub_ip"] = host["ip"]
 
-                host_custom_attrs = host.get("pytest_multihost")
+                host_custom_attrs = deepcopy(host.get("pytest_multihost", {}))
 
                 rm_keys = [key for key in host.keys() if key not in host_allowed_attrs]
                 for key in rm_keys:
                     del host[key]
 
-                if isinstance(host_custom_attrs, dict):
+                if host_custom_attrs:
                     host.update(host_custom_attrs)
 
             domain_rm_keys = [
--- a/tests/unit/test_ansible_inventory.py
+++ b/tests/unit/test_ansible_inventory.py
@@ -227,3 +227,77 @@ class TestAnsibleInventory:
         assert "something_else" in srv2
         assert srv2["no_ca"] == "yes"
         assert srv2["something_else"] == "for_fun"
+
+    def test_domain_arbitrary_attrs(self):
+        """
+        Test that values defined in `ansible_inventory` dictionary in host part
+        of job metadata file gets into host attributes in generated ansible
+        inventory.
+        """
+        metadata = metadata_extra()
+        metadata["domains"][0]["ansible_inventory"] = {
+            "no_ca": "no",
+            "something_else": "not_funny",
+        }
+
+        config = provisioning_config()
+        db = get_db_from_metadata(metadata)
+        ans_inv = AnsibleInventoryOutput(config, db, metadata)
+        inventory = ans_inv.create_inventory()
+
+        srv1 = inventory["all"]["hosts"]["srv1.example.test"]
+
+        assert "readonly_dc" not in srv1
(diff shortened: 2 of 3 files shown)


src/mrack/outputs/pytest_multihost.py
```diff
@@ -125,14 +125,18 @@ def create_multihost_config(self):  # pylint: disable=too-many-branches
                     mhcfg["ad_sub_hostname"] = host["name"]
                     mhcfg["ad_sub_ip"] = host["ip"]
 
-                host_custom_attrs = host.get("pytest_multihost")
+                glob_custom_attrs = self._metadata.get("pytest_multihost", {})
+                domain_custom_attrs = domain.get("pytest_multihost", {})
```
jrisc, line 129: Trivial detail, but here "global" is abbreviated  to `glob`, while "domain" is not. However, "domain" was abbreviated to `dom` in `ansible_inventory.py`. It would be more coherent to use the same abbreviation here.

tests/unit/test_ansible_inventory.py
```diff
@@ -227,3 +227,189 @@ def test_arbitrary_attrs(self):
...
+
+        srv1 = inventory["all"]["hosts"]["srv1.example.test"]
+
+        assert "readonly_dc" not in srv1
+        assert "something_else" in srv1
+        assert srv1["something_else"] == "not_funny"
+        assert "no_ca" in srv1
+        assert srv1["no_ca"] == "yes"
+
+        srv2 = inventory["all"]["hosts"]["srv2.example.test"]
```
jrisc, line 365: I think it would be nice to run this test with 2 domains, one host in each of them. Test that global scope attributes are set on the host (when they are not defined at domain level), and test that domain scope attributes have higher precedence than global scope ones.

I.e. Define an attribute at global level, and have 2 domains with one of them overwriting the global scope one.

jrisc (commented):
Tested with multiple metadata files. Inventory entries are properly forwarded to hosts with expected overriding behavior (the most specific has priority).

src/mrack/outputs/pytest_multihost.py
Tiboris: fixed

tests/unit/test_ansible_inventory.py
Tiboris: I will add this particular test case

Tiboris: Test added

jrisc (approved):
It looks good to me.

pvoborni (commented):
The Ansible part makes sense to me. 

But I'm not sure if we should have the pytest-multihost part done. Or done this way. The difference is that for Ansible, the config is set in-the end per-host thus merging from the layers above makes sense. 

But the pytest-multihost behaves differently. It itself  has multiple layers. Actually the job metadata format is basically pytest-multihost format with added "phases" section.  Meaning, that setting additional values makes sense on each of the 3 levels (root/global, domain, host). 

Right now, any value added on root level is part of the pytest multihost config.  The custom attrs were introduce to not blindly copy attrs from individual hosts because pytest-multihost has checks for attributes and fails if there is something which was not explicitly allowed by pytest-multihost or tested project. 

This implementation might confuse people e.g. thinking that they are setting something from the domain but infact it will get propagated to host.

Tiboris:
I can easily remove the multihost part if that would help to ship the ansible part required by users :innocent:

Tiboris:
Removed multihost changes

pvoborni (approved):
Thanks for the tests changes. This way it was much easier to review.

Tiboris:
Thanks for review :)

Issue #237, opened 2023-02-13
Issue: fqdn in name is ignored and mrack guesses the name instead


jakub-vavra-cz:
When the name is fqdn it is ignored and mrack creates name by getting the first part of the name and then add domain instead of using it directly.

https://github.com/neoave/mrack/blob/main/src/mrack/outputs/ansible_inventory.py#L146

pvoborni:
To clarify what is the expected issue and behavior - as the issue description doesn't provide it explicitly.

Atm, the current behavior is the following, right?

| Input: Host's name   | Input: Host's domain name | Output: meta_hostname | Output: meta_domain | Output meta_fqdn     | comment    |
| ---                  | ---                       | ---                   | ---                 | ---                  | ---        |
| abc.example.test     | example.test              | abc                   | example.test        | abc.example.test     | looks good |
| abc                  | example.test              | abc                   | example.test        | abc.example.test     | looks good |
| abc.sub.example.test | sub.example.test          | abc                   | sub.example.test    | abc.sub.example.test | looks good |
| abc                  | sub.example.test          | abc                   | sub.example.test    | abc.sub.example.test | looks good |
| abc.sub              | example.test              | abc                   | example.test        | abc.example.test     | the reported "issue" - "sub" is lost |

So the reported issue is the last case. I.e. mix of shortname and subdomain name in host's name and a parent domain in the domain part.

Is the expected behavior of the fix to be the following?

| Input: Host's name   | Input: Host's domain name | Output: meta_hostname | Output: meta_domain | Output meta_fqdn     |
| ---                  | ---                       | ---                   | ---                 | ---                  |
| abc.sub              | example.test              | abc                   | sub.example.test    | abc.sub.example.test |

jakub-vavra-cz:
When there is "." in the name mrack should recognize it as a fqdn and just return it.
Otherwise it should append the domain to the name. 
#238

pvoborni:
From my PoV, mrack should rather fail than to this.  Such behavior is even more confusing. Somebody else would open a bug exactly about this - saying: why is the FQDN of the host not containing the domain where the host is in.

pvoborni:
What also looks correct to me is:

Fail on validation if hostname is e.g.  "abc.foo" and domain is "example.test" - i.e. if host tries to be FQDN but doesn't contain it's parent domain name.

Support the following case:

[...]

PR #238, opened 2023-02-13
fix: fqdn in name is ignored and mrack guesses the name instead #237

jakub-vavra-cz:


tests/unit/test_utils.py
```diff
@@ -16,6 +16,17 @@ class TestPytestMrackUtils:
     def test_get_shortname(self, hostname, expected):
         assert get_shortname(hostname) == expected
 
+    @pytest.mark.parametrize(
+        "hostname,domain,expected",
+        [
+            ("abc.test", "bar", "abc.test"),
```
pvoborni, line 22: A host under domain `bar` cannot have a FQDN not containing the domain `bar`. In this case the the `abc.test` must be either under TLD domain `test` or the FQDN should be e.g. `abc.test.bar`

src/mrack/outputs/ansible_inventory.py
```diff
@@ -143,7 +144,7 @@ def create_ansible_host(self, name):
             "ansible_host": ansible_host,
             "ansible_python_interpreter": python,
             "ansible_user": ansible_user,
-            "meta_fqdn": f"{get_shortname(name)}.{dom_name}",
+            "meta_fqdn": f"{get_fqdn(name, dom_name)}",
```
pvoborni, line 147: It must be ensured that `meta_fqdn` = `meta_hostname` + `meta_domain`.

pvoborni (changes requested):
Hi, this patch is breaking some behaviors expected from mrack.  

I think it would be good to agree first on what is the expected behavior - e.g. in the issue description.

tests/unit/test_utils.py
jakub-vavra-cz: Point of this test is that when name is fqdn, domain is ignored/overridden by whatever is in the hostname.
**bar** is there only to make it clear.

src/mrack/outputs/ansible_inventory.py
jakub-vavra-cz: The intent here is to be able to override this assumption and have hosts from subdomains directly in the same tld domain by specifying names as fqdn.
```
domains:
  - name: test
    type: sssd
    hosts:
      - name: client.test
        group: medium
        role: client
        os: rhel-9.2
      - name: dns.test
        group: xsmall
        role: dns
        os: rhel-9.2
      - name: master.ldap.test
        group: medium
        role: ldap
        os: rhel-9.2
      - name: master.ipa.test
        group: medium
        role: ipa
        os: rhel-9.2
```

Tiboris: i would expect domain name to be: something.com

pvoborni: To keep discussion on one place, let's continue the discussion in the issue: https://github.com/neoave/mrack/issues/237

jakub-vavra-cz: Resolved.

tests/unit/test_utils.py
jakub-vavra-cz: Resolved.

pvoborni (approved):
This behavior looks good to me. Tests looks good as well. Thanks.

 Maybe @Tiboris  can look at it as well before we merge.

There might also some further improvements done, but they don't block the MR (can be done separately):

- fail early with the error, i.e. before we start provision and not at the end. This could be achieved by e.g. extending https://github.com/neoave/mrack/blob/main/src/mrack/actions/up.py#L64 
- the method passes even if host's name == domain name :)  - it's quite a corner case so it doesn't worry me much

Tiboris (approved):
Thanks for the patch

PR #242, opened 2023-02-20
feat: Add pytest-mh output and "config" section to metadata

pbrezina:
This adds new output for pytest-mh configuration file. Pytest-mh is
a complete rewrite of pytest-multihost with additional features that
is going to be used by SSSD. See: https://pytest-mh.readthedocs.io

This PR also adds `config` section to metadata file that currently can
contain:

* `outputs`: list of output files, defaults to `"anisble-inventory", "pytest-multihost`
* `ansible/layout`: to overwrite inventory_layout from provisioning-config

### DIFF
--- a/src/mrack/outputs/ansible_inventory.py
+++ b/src/mrack/outputs/ansible_inventory.py
@@ -23,6 +23,7 @@ from mrack.outputs.utils import get_external_id
 from mrack.utils import (
     get_fqdn,
     get_host_from_metadata,
+    get_os_type,
     get_password,
     get_shortname,
     get_ssh_key,
@@ -144,6 +145,7 @@ class AnsibleInventoryOutput:
             "ansible_host": ansible_host,
             "ansible_python_interpreter": python,
             "ansible_user": ansible_user,
+            "meta_os_type": get_os_type(meta_host),
             "meta_fqdn": f"{get_fqdn(name, dom_name)}",
             "meta_hostname": get_shortname(name),
             "meta_domain": f"{get_fqdn(name, dom_name).split('.', 1)[1]}",
--- a/src/mrack/outputs/ansible_inventory.py
+++ b/src/mrack/outputs/ansible_inventory.py
@@ -74,7 +74,7 @@ def get_group(inventory, groupname):
         else:
             if "children" in group:
                 found = get_group(group["children"], groupname)
-        if found:
+        if found is not None:
             break
     return found
 
@@ -86,7 +86,7 @@ def add_to_group(inventory, groupname, hostname):
     Returns group or None if it doesn't exist.
     """
     group = get_group(inventory, groupname)
-    if not group:
+    if group is None:
         return None
 
     if "hosts" not in group:
--- a/src/mrack/outputs/ansible_inventory.py
+++ b/src/mrack/outputs/ansible_inventory.py
@@ -202,12 +202,18 @@ class AnsibleInventoryOutput:
 
         return host_info
 
+    def get_inventory_layout(self):
+        """Get Ansible inventory layout."""
+        layout = self._metadata.get("config", {}).get("ansible", {}).get("layout")
+        if layout is None:
+            return self._config.get("inventory_layout", DEFAULT_INVENTORY_LAYOUT)
+
+        return layout
+
     def create_inventory(self):
         """Create the Ansible inventory in dict form."""
         provisioned = self._db.hosts
-        inventory = deepcopy(
-            self._config.get("inventory_layout", DEFAULT_INVENTORY_LAYOUT)
-        )
+        inventory = deepcopy(self.get_inventory_layout())
         if not isinstance(inventory, dict):
             raise ConfigError("Inventory layout should be a dictionary")
         all_group = ensure_all_group(inventory)
--- a/src/mrack/outputs/pytest_mh.py
+++ b/src/mrack/outputs/pytest_mh.py
@@ -1,0 +1,88 @@
+# Copyright 2023 Red Hat Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""pytest-mh configuration file output module."""
+
+import logging
+
+from mrack.outputs.utils import get_external_id
+from mrack.utils import get_fqdn, get_os_type, save_yaml
+
+DEFAULT_MHCFG_PATH = "pytest-mh.yaml"
+
+logger = logging.getLogger(__name__)
+
+
+class PytestMhOutput:
+    """
+    Create a configuration file for pytest-mh.
+
+    From job metadata and provisioned information (DB).
+
+    Structure of mh config file is almost the same as our topology in job
+    metadata definition.
+    """
+
+    def __init__(self, config, db, metadata, path=None):
+        """Init the output module."""
+        self._config = config
+        self._db = db
+        self._metadata = metadata
+        self._path = path or DEFAULT_MHCFG_PATH
+
+    def create_mh_config(self):
+        """
+        Create configuration file for python-pytest-multihost.
+
+        Return in form of dict.
+        """
+        cfg = {"domains": []}
+        for domain in self._metadata.get("domains", []):
+            cfgdom = {"id": domain["name"], "hosts": []}
+            cfg["domains"].append(cfgdom)
+
+            for host in domain.get("hosts", []):
+                if "role" not in host:
+                    continue
+
+                provisioned_host = self._db.hosts.get(host["name"], None)
+                if not provisioned_host:
+                    logger.error(f"Host {host['name']} not found in the database.")
+                    continue
+
+                cfgdom["hosts"].append(
+                    {
+                        "hostname": get_fqdn(host["name"], domain.get("name", "")),
+                        "os": {
+                            "family": get_os_type(host),
+                        },
+                        "role": host["role"],
+                        "ssh": {
+                            "host": get_external_id(
+                                provisioned_host, host, self._config
+                            ),
+                        },
+                        **host.get("pytest_mh", {}),
+                    }
+                )
+
+        return cfg
+
+    def create_output(self):
+        """Create the target output file."""
+        mhcfg = self.create_mh_config()
+        save_yaml(self._path, mhcfg, sort_keys=False)
+        if self._path:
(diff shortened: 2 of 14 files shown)


pbrezina:
The test pass locally fine. I do not know why it fails and the log does not provide any clue. Are you able to reproduce it?

Tiboris:
Hello @pbrezina, thanks for the patches! I would like to kindly ask you to change your commit messages so they are valid for our automation see run https://github.com/neoave/mrack/actions/runs/4223470841/jobs/7353482982  

Currently we support (see error message):
```
Error: Error: Commit message should match the '<type>(<scope>): <subject>' style 
 for python-semantic-release supported types:
 feat: A new feature
 fix: A bug fix
 docs: Documentation only changes
 style: Changes that do not affect the meaning of the code (white-space, formatting, missing chars...)
 refactor: A code change that neither fixes a bug nor adds a feature
 perf: A code change that improves performance
 test: Adding missing or correcting existing tests
 chore: Changes to the build process or auxiliary tools and libraries such as documentation generation
```

so in you case eg: refactor(black) ; feat(outputs) etc.

pbrezina:
Done.

Tiboris:
Could you please rebase ? I will try to pull the code and run tests locally.

jakub-vavra-cz (changes requested):

src/mrack/outputs/ansible_inventory.py
```diff
@@ -200,12 +200,18 @@ def create_ansible_host(self, name):
 
         return host_info
 
+    def get_inventory_layout(self):
```
jakub-vavra-cz, line 203: I do not see much point in adding a custom inventory layout, we can add host groups using "groups:" already.

pvoborni: Inventory layout purpose is not adding groups but rather grouping of groups, creating a hierarchy. This is useful in general preparation playbooks where plays can be limited to e.g. linux or windows. Or some other plays can be limited to very special groups.

src/mrack/outputs/pytest_mh.py
```diff
@@ -0,0 +1,86 @@
...
+
+                provisioned_host = self._db.hosts.get(host["name"], None)
+                if not provisioned_host:
+                    logger.error(f"Host {host['name']} not found in the database.")
+                    continue
+
+                cfgdom["hosts"].append(
+                    {
+                        "hostname": get_fqdn(host["name"], domain.get("name", "")),
+                        "os": "linux" if not is_windows_host(host) else "windows",
```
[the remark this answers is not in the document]
pbrezina: IMHO host_type is semantically incorrect and it is not present in metadata unless explicitly added, this attribute is used only by pytest_multihost. Mrack uses os_type inside is_windows_host but it is also not defined implicitly. We can name it os_type if you prefer.

src/mrack/outputs/ansible_inventory.py
pbrezina: If we change the sssd-ci-containers playbooks, then it is not that important to stick to particular layout. However, being able to override the default values and clear them is still important so things don't get broken by changing the value in provisioning-config.

pbrezina:
Rebased.

src/mrack/outputs/pytest_mh.py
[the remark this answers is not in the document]
pvoborni: Given that mrack already has the  `os_type` as one of the methods for figuring out if it is windows or not. Then propagating this else where makes sense to me. E.g. we can also, in different PR, enhance pytest-multihost output to translate `os_type` to `host_type` (but still support `host_type`) so that usage of `os_type` could be common for this kind of behavior for all the outputs.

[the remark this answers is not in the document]
pbrezina: Do you set os_type somewhere outside metadata? It is not set in my case so either I can enforce it being set or I need to stick with is_windows_host() anyway.

src/mrack/outputs/ansible_inventory.py
jakub-vavra-cz: I am afraid that we will have a huge definition of layout copypasted in metadata making it long and messy.
Maybe adding an option to have the layout in external file would be better.
The config then can have just one line with path referring to the layout/template.

src/mrack/outputs/pytest_mh.py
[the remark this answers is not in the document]
pvoborni: Atm no. 

But it makes sense to me to add it ansible inventory host vars.  Might be useful for some decisions in playbooks.

Something in spirit of:
```
# utils.py
def get_os_type(meta_host).
    os_type = meta_host.get("os_type" )
    if not os_type:
        "os_type" =  "linux" if not is_windows_host(meta_host) else "windows",   

# ansible inventory output
inventory_host["meta_os_type"] = get_os_type(host)
```

src/mrack/outputs/ansible_inventory.py
pvoborni: Not against, but it is not that far from what can be done now: kinda in similar spirit (overriding some file on defined location). You can now add a step (playbook, command or something else) in metadata after the `playbook: init/testrunner.yaml`. It can load the provisioning config, modify the layout and save the file.  Yes, it is slightly more complex than just replacing a file, but actually not that much as you still need some step to do it.

Nowdays, I'm more thinking about generic way how to extend + override provisioning config - some kind of clever merge. There might be generic stuff, useful for multiple jobs defined e.g. on upstream, downstream and project specific level.

pbrezina: There is https://github.com/tanbro/pyyaml-include that gives you include for free.

jakub-vavra-cz: BTW: I got bitten in my ass by the magic done with the inventory twice now so I vote for extracting the layout from provisioning config to standalone file that can be included.
1. I used group ipa in my inventory and have machines marked a ipaserver and ipaclient (to select flavour) added to it unexpectedly.
2. A new group client was added resulting in name collision with a machine named client.

We could have layouts for different projects pre-defined in different files and have mrack-up with extra_vars to select from them or add a custom one.

src/mrack/outputs/pytest_mh.py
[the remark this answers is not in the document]
pbrezina: Ok, I added get_os_type to mrack and renamed python-mh config option to "os_family" to make it clear what it means.

src/mrack/outputs/ansible_inventory.py
pbrezina: I think the way to go would be to allow override everything from provisioning-config in metadata through the config section (both merge and replace mechanisms should imho be supported) as we talked about in the office. Include makes sense to me and can be easily implemented by pyyaml-include. But I don't see any work related to this PR so I'm marking this as resolved. Please, reopen it if you think otherwise.

src/mrack/outputs/pytest_mh.py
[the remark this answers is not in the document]
pbrezina: Hopefully, this comment is now resolved.

tests/unit/test_pytest_mh.py
```diff
@@ -0,0 +1,87 @@
...
+        expected = yaml.safe_load(
+            """
+        domains:
+        - type: sssd
+          hosts:
+          - hostname: client.test
+            os: linux
+            role: client
+            ssh:
+              host: _gateway
```
pvoborni, line 69: Any reason to be `_gateway` ? In my local setting the test works if both`_gateway` are replaced by `192.168.0.1` - created in `create_db_host` method. This should solve the CI issue.

```
diff --git a/tests/unit/test_pytest_mh.py b/tests/unit/test_pytest_mh.py
index 39e3e3e..89a60d9 100644
--- a/tests/unit/test_pytest_mh.py
+++ b/tests/unit/test_pytest_mh.py
@@ -66,7 +66,7 @@ class TestPytestMhOutput:
             os: linux
             role: client
             ssh:
-              host: _gateway
+              host: 192.168.0.1
             artifacts:
             - /etc/sssd/*
             - /var/log/sssd/*
@@ -75,7 +75,7 @@ class TestPytestMhOutput:
             os: linux
             role: ipa
             ssh:
-              host: _gateway
+              host: 192.168.0.1
             config:
               client:
                 ipa_domain: ipa.test
(.env) pvoborni@ntb4 ~/d/n/mrack ((76c20d15))> .env/bin/pytest tests/unit/test_pytest_mh.py
========================================= test session starts ==========================================
platform linux -- Python 3.11.1, pytest-7.2.1, pluggy-1.0.0
rootdir: /home/pvoborni/dev/neoave/mrack, configfile: pytest.ini
plugins: asyncio-0.20.3
asyncio: mode=Mode.STRICT
collected 1 item                                                                                       

tests/unit/test_pytest_mh.py .                                                                   [100%]

========================================== 1 passed in 0.34s ===========================================
```

tests/integration/test_actions.py
```diff
@@ -62,27 +62,55 @@ async def test_list_action(self, database, setup_providers, caplog):
...
     @pytest.mark.asyncio
+    @pytest.mark.parametrize(
+        "config,files",
+        [
+            (None, ["mrack-inventory.yaml", "pytest-multihost.yaml"]),
+            (
+                ["ansible-inventory", "pytest-mh"],
+                ["mrack-inventory.yaml", "pytest-mh.yaml"],
+            ),
+            (
```
pvoborni, line 73: This test looks as duplicate of previous. Did you meant to use `pytest-multihost` instead of `pytest-mh` ?

pvoborni (changes requested):
Finally found time to read it whole and do some minor testing. 

There are 2 minor issues in tests which needs fixing but otherwise looks good to me.  

Thanks for the contribution.

tests/integration/test_actions.py
pbrezina: Yes.

tests/unit/test_pytest_mh.py
pbrezina: _gateway is present in my local test, if I provide the IP address the test fails. But it works in PRCI. Fixed.

pbrezina:
Please, don't push it yet. I will probably also add the domain name into the configuration.

src/mrack/outputs/pytest_mh.py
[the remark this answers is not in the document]
jakub-vavra-cz: LGTM.

pbrezina:
@pvoborni It is ready for final review. I change the "type" in pytest-mh to "id" (taken from domain["name"]). And I needed to relax condition in get_fqdn so the domain does not have to match. I assume it is ok, I can avoid using get_fqdn if not.

Tiboris (approved):
Regression tests passing, if it is ready, we are ready.

pbrezina:
I amended https://github.com/neoave/mrack/pull/242/commits/dcaf0bc99704a773a062f63ceba275dd9e2a60c3 to catch also case when there are more nested groups. In this case, it would only successfully find the last one. Line 77 was changed and test was extended.

pbrezina:
Now it is ready to push.

PR #243, opened 2023-02-21
fix(OpenStack/Beaker): validation fails with exception when image/distro not found

Tiboris:
fix(OpenStack): Add exception parameter when validation fails
    
    Adding second parameter to ValidationError exception gets rid of mrack exiting with exception.

fix(Beaker): rerurn common dictionary when validation fails

    when Beaker distro/image is not recognized by beakerhub
    mrack failed with exception as newer code expects
    dictionary error object.

fix: Use get method when host error object is a dictionary

    This makes sure to not throw an exception when host.error
    is not a dictionary and we can not use .get() method

### DIFF
--- a/src/mrack/providers/beaker.py
+++ b/src/mrack/providers/beaker.py
@@ -47,7 +47,7 @@ PROVISIONER_KEY = "beaker"
 
 
 def parse_bkr_exc_str(exc_str):
-    """Parse exception string and return more readable string for mrack error."""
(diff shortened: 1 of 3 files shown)


kaleemsiddiqu (approved):
Tested locally

    2023-02-22 12:11:38,566 mrack.actions.up ERROR OpenStack Image not found image: rhel-9.3, ref: None
    2023-02-22 12:11:38,570 mrack.run ERROR Provider(s) OpenStack failed to provision resources

PR #245, opened 2023-02-22
fix(OpenStack): load limits properly by one method

Tiboris:
when creatring utilization method for openstack there was a problem that limits were read incorrectly.
moving code to private method reused in utilization and can_provision methods to get rid of copy-paste errors.


dav-pascual (approved):
Thanks for the patch @Tiboris 
LGTM

Issue #248, opened 2023-03-01
OpenStack fails to provision in v1.13.2 


Tiboris:
```
  File "/home/tdudlak/Work/mrack/src/mrack/run.py", line 151, in up
    await up_action.provision()
  File "/home/tdudlak/Work/mrack/src/mrack/actions/up.py", line 106, in provision
    raise results
  File "/home/tdudlak/Work/mrack/src/mrack/providers/provider.py", line 402, in provision_hosts
    success_hosts, error_hosts, _missing_reqs = await self.strategy_retry(reqs)
                                                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/tdudlak/Work/mrack/src/mrack/providers/provider.py", line 448, in strategy_retry
    s_hosts, error_hosts, missing_reqs = await self._provision_base(
                                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/tdudlak/Work/mrack/src/mrack/providers/provider.py", line 280, in _provision_base
    while not await self.can_provision(reqs):
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/tdudlak/Work/mrack/src/mrack/providers/openstack.py", line 732, in can_provision
    
TypeError: cannot unpack non-iterable coroutine object
```
its caused by:
```python
used_vcpus, used_memory, limit_vcpus, limit_memory = self._load_limits()
```
not being awaited

PR #249, opened 2023-03-01
fix(OpenStack): await loading limits to not break provisioning

Tiboris:
Resolves: https://github.com/neoave/mrack/issues/248


Tiboris:
Tested with internal pipeline and confirmed with @dav-pascual it fixes the issue

dav-pascual (approved):
LGTM

PR #251, opened 2023-03-07
fix(OpenStack): Add missing await for self._load_limits() method call

Tiboris:
one more await is missing for this method so fixing it.


dav-pascual (approved):
Thanks for the patch @Tiboris !

PR #253, opened 2023-03-10
fix: Replace coroutines with tasks to avoid RuntimeError

dav-pascual:
Fixing code block in `_opentack_gather_responses` method to avoid RuntimeError when a coroutine is reused in the loop.

Including a unit test for this method as part of the commit.

The issue was that a coroutine (aka the async function calls parameters) cannot be reused with await or gather, hence when a ServerError exception was raised, it caused a RunTime error. A workarround for this is awaiting using tasks instead of couroutines directly, and recreate the tasks if necessary to call the coroutines again.

### DIFF
--- a/tests/unit/test_openstack.py
+++ b/tests/unit/test_openstack.py
@@ -11,15 +11,22 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+import asyncio
 import os
 from copy import deepcopy
 from unittest import mock
 from unittest.mock import Mock, patch
 
 import pytest
+from simple_rest_client.exceptions import ServerError
 
 from mrack.context import global_context
-from mrack.errors import MrackError, ProviderNotExists, ValidationError
+from mrack.errors import (
+    MrackError,
+    ProviderNotExists,
+    ProvisioningError,
+    ValidationError,
+)
 from mrack.providers.openstack import OpenStackProvider
 
 from .mock_networks import (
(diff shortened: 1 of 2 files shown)


pvoborni:
I think this patch doesn't solve the issue. The bug seems to be introduced in https://github.com/neoave/mrack/commit/8ef831297a99c47eb270a6ea4f5e33ed20c8caee The regression is that on re-try the functions are not called again. 

The whole idea is that on server error we call openstack again to get the data to try to avoid the ServerError.

dav-pascual:
> I think this patch doesn't solve the issue. The bug seems to be introduced in [8ef8312](https://github.com/neoave/mrack/commit/8ef831297a99c47eb270a6ea4f5e33ed20c8caee) The regression is that on re-try the functions are not called again.
> 
> The whole idea is that on server error we call openstack again to get the data to try to avoid the ServerError.

@pvoborni 
I will review this further with @Tiboris 
but, what was preventing to re-call the functions was the incorrect use of `asyncio.gather` which cannot be reused with coroutines, hence provoking the RunTime error after the first failed iteration of the while loop and preventing the retry.

pvoborni:
I don't think I was referring to that.

This method is called in e.g.:
```python
_, _, self.limits, _, _ = await self._openstack_gather_responses(
            self._load_flavors(),
            self._load_images(image_names),
            self.nova.limits.show(),
            self._load_networks(),
            self._load_ip_availabilities(),
        )
```
On server error we need to run again the methods. Resp. it would be good to re-run only the failed one., e.g. `self._load_images(image_names)`.  Wrapping a coroutine with task will not accomplish the re-run of the function call which created the coroutine. We probably need to pass the functions and their params to `_openstack_gather_responses` so that inside it can be re-run again. 

We can also say that the created test is not testing the scenario correctly. There is no re-run of initial function. Proper test would have e.g. 2 calls of `coro2` inside the `_openstack_gather_responses` where one would end with `ServerError` and second would succeed thus the function would return `result[2]` with proper result.

tests/unit/test_openstack.py
```diff
@@ -594,3 +601,44 @@ async def test_network_picking_no_spread(self, x):
...
+        provider = OpenStackProvider()
+
+        # Test successful outcome
+        result = await provider._openstack_gather_responses([coro1(), [], {}])
+        assert result[0] == "result1"
+
+        # Test that we get successful outcome after getting a ServerError
+        error_flag = True
+        result = await provider._openstack_gather_responses(
+            [coro1(), [], {}], [coro2(), [], {}]
```
Tiboris, line 635: I think you forgot brackets here
```suggestion
            [coro1, [], {}], [coro2, [], {}]
```

pvoborni (approved):
Looks good to me and provisioning worked.

tests/unit/test_openstack.py
dav-pascual: thanks, fixed

Tiboris (approved):
Thanks @dav-pascual for the patch and @pvoborni for knowledge sharing session.

PR #254, opened 2023-03-16
fix: empty the success hosts and correctly set missing_reqs

Tiboris:
Since I have coded a regression when trying to re-provision
mrack is not setting variables for another loop correctly
error_hosts is updated correctly but when keeping success_hosts
not changed to empty list it creates a illusion of hosts
which were deleted because of re-provisioning strategy
that they are still up and running.

### DIFF
--- a/src/mrack/providers/provider.py
+++ b/src/mrack/providers/provider.py
@@ -376,11 +376,7 @@ class Provider:
         else:  # we do not check the ssh connection to VMs
             success_hosts = active_hosts
 
-        missing_reqs = [
-            req for req in reqs if req["name"] in [host.name for host in error_hosts]
-        ]
-
-        return (success_hosts, error_hosts, missing_reqs)
+        return (success_hosts, error_hosts, self._get_missing_reqs(reqs, error_hosts))
 
     async def provision_hosts(self, reqs):
         """Provision hosts based on list of host requirements.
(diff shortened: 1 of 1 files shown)


src/mrack/providers/provider.py
```diff
@@ -413,6 +409,12 @@ async def provision_hosts(self, reqs):
 
         return success_hosts
 
+    def _get_missing_reqs(self, reqs, error_hosts):
+        """Rerurn missing requirements based on reqs and error hosts."""
```
pvoborni, line 413: Typo in the description.

pvoborni (changes requested):
The code looks good to me. But it is hard to tell whether it indeed works.  This kind of scenario is quite difficult to reproduce in adhoc testing. A test with good usage of OpenStack calls mocking which would simulate the scenario would be great to have.

Also the commit/MR description doesn't really describe the scenario which is not working.

src/mrack/providers/provider.py
Tiboris: fixed and updated commit message.

pvoborni (approved):
I'm approving this as the commit has better message and I saw that the PR really helps in internal real testing. 

That said we still need to deliver the tests soon, but not in this PR.

PR #255, opened 2023-03-28
fix: Handle 403 AuthError (out of quota) in openstack provisioning

dav-pascual:
This modification will catch the Exception, to free all hosts and wait for quota to be available again

### DIFF
--- a/tests/unit/test_openstack.py
+++ b/tests/unit/test_openstack.py
@@ -11,7 +11,6 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-import asyncio
 import os
 from copy import deepcopy
 from unittest import mock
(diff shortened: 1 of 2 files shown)


Tiboris (approved):
Thanks @dav-pascual we weng through the changes together and I hope we catch it properly here. Later the resource check and error host creation will make sure resources are freed correctly.

Tiboris:
I believe we can try to set same approach as in https://github.com/neoave/mrack/pull/254 and deliver tests later or create automation to run some tests internally.

pvoborni:
We should implement the tests and run them before we release next version.  I don't want to see extended usage of internal tests.  Better practice is having them with code, running on the changes, catching issues immediately.  Also you cannot reproduce these scenarios in production consistently so internal tests are not the solution here.

dav-pascual:
After further discussion, I will develop a unit test for create_server, while we look forward to close the gaps in the future regarding the untested code

dav-pascual:
Updated with new unit test.
Aditionally refactoring _openstack_gather_responses to fix test warnings "coroutine was never awaited" and improving test exec time by mocking async sleeps calls.

tests/unit/test_openstack.py
```diff
@@ -603,7 +603,8 @@ async def test_network_picking_no_spread(self, x):
         )
 
     @pytest.mark.asyncio
-    async def test_openstack_gather_responses(self):
+    @patch("asyncio.sleep", new_callable=AsyncMock)
```
pvoborni, line 606: I guess the patch of `asyncio.sleep` is done because of the usage in `._openstack_gather_responses. But it also makes its calls in `coro1`, `coro2`, `coro3` not doing anything so the `asyncio.sleep` can be removed from those coros.

pvoborni (changes requested):
Thanks for the tests. Mostly looks good to me. There are only minor things.

pvoborni (approved):

Tiboris (approved):
Thanks @dav-pascual

PR #256, opened 2023-03-31
feat(outputs): merge nested dictionary instead of overriding it

pbrezina:
pytest-mh configuration may contain nested dictionary, specifying
username or password in pytest_mh section will result in overriding
the host option, i.e.:

```
pytest_mh:
  ssh:
    username: tuser
    password: tuserpassword
```

will generate wrong configuration where ssh/host is not set.

This patch makes sure that the ssh key is merged instead of overwritten.

### DIFF
--- a/src/mrack/outputs/pytest_mh.py
+++ b/src/mrack/outputs/pytest_mh.py
@@ -17,7 +17,7 @@
 import logging
 
 from mrack.outputs.utils import get_external_id, merge_dict
-from mrack.utils import get_fqdn, get_os_type, save_yaml
+from mrack.utils import get_fqdn, get_os_type, get_password, get_username, save_yaml
 
 DEFAULT_MHCFG_PATH = "pytest-mh.yaml"
 
@@ -71,6 +71,15 @@ class PytestMhOutput:
                         "host": get_external_id(provisioned_host, host, self._config),
                     },
                 }
+
+                if get_os_type(host) == "windows":
+                    username = get_username(provisioned_host, host, self._config)
+                    if username:
+                        hostcfg["ssh"]["username"] = username
+
+                    password = get_password(provisioned_host, host, self._config)
+                    if password:
+                        hostcfg["ssh"]["password"] = password
 
                 cfgdom["hosts"].append(merge_dict(hostcfg, host.get("pytest_mh", {})))
 
--- a/src/mrack/outputs/pytest_mh.py
+++ b/src/mrack/outputs/pytest_mh.py
@@ -16,7 +16,7 @@
 
 import logging
 
-from mrack.outputs.utils import get_external_id
+from mrack.outputs.utils import get_external_id, merge_dict
 from mrack.utils import get_fqdn, get_os_type, save_yaml
 
 DEFAULT_MHCFG_PATH = "pytest-mh.yaml"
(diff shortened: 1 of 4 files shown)


src/mrack/outputs/pytest_mh.py
```diff
@@ -61,21 +63,19 @@ def create_mh_config(self):
...
-                        **host.get("pytest_mh", {}),
-                    }
-                )
+                hostcfg = {
+                    "hostname": get_fqdn(host["name"], domain.get("name", "")),
+                    "os": {
+                        "family": get_os_type(host),
+                    },
+                    "role": host["role"],
+                    "ssh": {
```
jakub-vavra-cz, line 72: Hi there should be used mrack function get_username(if the host is a windows machine.) and get_password.  see https://github.com/neoave/mrack/blob/main/src/mrack/outputs/pytest_multihost.py#L103

jakub-vavra-cz (changes requested):
Please make sure that the password and username handling aligns with multihost output.

src/mrack/outputs/pytest_mh.py
pbrezina: Done. I'm not sure how to write unit test for it. @pvoborni is there a way to mock windows machine that would hit this code path?

pbrezina:
> The packit tests fails as the spec file is missing the new python dependency. But even if added, I think it will fail on it missing in epel.

Ok, I removed the dependency and implement the function from scratch.

src/mrack/outputs/utils.py
```diff
@@ -45,3 +46,19 @@ def get_external_id(host, meta_host, config):
     if resolve_ip:
         external_id = resolve_hostname(host.ip_addr) or host.ip_addr
     return external_id
+
+
+def merge_dict(d1, d2):
```
Tiboris, line 51: May I ask if the simple operator `|` would not solve your issue?
```
In [1]: a = {"A" : 1, "B": 2}

In [2]: b = {"B": 200, "C": 3}

In [3]: a | b
Out[3]: {'A': 1, 'B': 200, 'C': 3}

In [4]: b| a
Out[4]: {'B': 2, 'C': 3, 'A': 1}

In [5]:
```

src/mrack/outputs/pytest_mh.py
pvoborni: Set `os` to  startwith `win` or set `os_type` to `windows` for some of the hosts in mocked job metadata in the test.

pvoborni: This can be added by provisioning provider, e.g. if the cloud environment generates the password. But I don't understand why you need it.  You can also set the password on job metadata level for the host.

pvoborni: That said. The first 'host' is the DB host. So if the test is mocking the DB then it is the thing to set.

Tiboris:
@pbrezina could you please re-push all three commits with different hashes as TF thinks there still old change there?

jakub-vavra-cz (approved):
LGTM.

pbrezina:
Done. I also added test for the windows case.

pbrezina:
Packit still tries to find "mergedeep" package but there is no reference to it in my commits...

pvoborni:
@Tiboris I'd approve and merge this PR even if the packit fails as it doesn't seem to be related. WDYT?

It looks like we need to improve packit config to include some kind of timestamp in package version so that newer build would be created+used by packit.

Tiboris (approved):
Thanks @pbrezina lets merge this!

PR #258, opened 2023-04-06
chore: fix docs dependencies in tox run

pvoborni:
Running `make test` or `tox` failed on building docs as the tox config did not use the docsi/requirements.txt and thus the required dependency which was recently added was missing.

This path is changing the tox run to use the docs/requirements.txt and thus is fixing the issue.


Tiboris (approved):
Thanks @pvoborni

PR #259, opened 2023-04-06
feat: configurable ssh options

pvoborni:
This implements configurable ssh options according to design:

https://mrack.readthedocs.io/en/latest/designs/ssh_options.html

With the idea that those options will be used for SSH connections check as well as added to generated Ansible inventory so that other tools can use it.


XiyangDong (commented):
LGTM, unit tests for testing ssh options are sophisticated and passing hence approving, but I'll leave it to Tibor for the final approval since this is mrack.

PR #260, opened 2023-04-14
test: Add missing tests for recent fixes and features

dav-pascual:
Specifics in commit

### DIFF
--- a/src/mrack/providers/provider.py
+++ b/src/mrack/providers/provider.py
@@ -424,6 +424,10 @@ class Provider:
             req for req in reqs if req["name"] in [host.name for host in error_hosts]
         ]
 
+    def _get_max_utilization(self):
+        """Get max utilization value from context."""
+        return global_context.CONFIG.max_utilization
+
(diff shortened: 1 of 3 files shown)


src/mrack/providers/provider.py
```diff
@@ -432,7 +436,7 @@ async def strategy_retry(self, reqs):
         success_hosts = []
         error_hosts = []
         delta_sleep = global_context.CONFIG.delta_sleep
-        max_utilization = global_context.CONFIG.max_utilization
+        max_utilization = self._get_max_utilization()
```
Tiboris, line 439: One nitpick, could you please extract this to another 'refactor' commit please?

Tiboris (changes requested):
One minor nitpick David, and I think we can go for merge. Do you agree?

src/mrack/providers/provider.py
dav-pascual: No problem, updated

Tiboris (approved):
Thanks for change!

PR #263, opened 2023-06-07
fix: Do not reprovision all hosts when server error is detected 

Tiboris:
when 500 code is detected in the error state of the errored VM
mrack would aggressively remove all hosts which have been
already provisioned properly. Changing logic that
if only a certain percentage of errored hosts is reached.

fix: Use lower cooldown time to not be too slow in re-provisioning


dav-pascual:
I think this looks good, thanks

Probably a testcase for the high share of failed hosts case would be nice. Should be quick modify one of the existing ones :)

Tiboris:
@dav-pascual Added test

dav-pascual (approved):

tests/unit/test_provider.py
```diff
@@ -195,6 +195,64 @@ async def subtest_success():
...
+            missing_dummy_reqs = [
+                {"name": "host3", "vcpus": 2, "memory": 4096},
+            ]
+
+            call_counts = [0]  # Initialize a list to store the call counts
+
+            def provision_side_effect(*args, **kwargs):
+
+                call_counts[0] += 1  # Increment the call count
+                print(f"CALL NO: {call_counts}\n{call_counts[0]}")
```
Tiboris, line 214: @dav-pascual I will remove these prints and repush, are commit messages fine? i mean especially the one with feat: does feat make sense there or should I rather use fix?

Tiboris: I have changed that to fix. it is not major change

PR #264, opened 2023-06-13
chore(Packit): Use yaml magic to run same internal tests for PRs and commits to main

Tiboris:
Use yaml magic to run same internal tests for PRs and commits to main


lachmanfrantisek (approved):
Looks good to me from the Packit side.

dav-pascual (approved):
Works for me

PR #266, opened 2023-07-21
Add clouds.yaml as an authentication method & Import publick key on provision

dav-pascual:
- feat(OpenStack): Import publick key on provision
This will save manual steps and error if the public key was not previously imported to
openstack profile.

- feat(OpenStack): Add clouds.yaml as an authentication method
Add clouds.yaml as an alternative to env vars.
Also, user+password and application credentials formats are both allowed.


src/mrack/providers/openstack.py
```diff
@@ -117,8 +121,102 @@ async def _openstack_gather_responses(self, *calls):
...
+
+                # Get the cloud specified in provisioning-config
+                # If that fails try to pick the cloud specified in OS_CLOUD envvar
+                # or get the only cloud (if there is only one)
+                try:
+                    cloud = config.get_one_cloud(self.cloud_profile)
+                except os_client_config.exceptions.OpenStackConfigException:
+                    cloud = config.get_one_cloud()
+
+                auth_info = cloud.config["auth"]
```
pvoborni, line 151: I'd add bit more logging here. E.g. when I was testing it, I did not know from the logs what credentials it actually picked so I had to use:

```diff
diff --git a/src/mrack/providers/openstack.py b/src/mrack/providers/openstack.py
index cd9613c..0d81d6e 100644
--- a/src/mrack/providers/openstack.py
+++ b/src/mrack/providers/openstack.py
@@ -149,8 +149,10 @@ class OpenStackProvider(Provider):
                     cloud = config.get_one_cloud()
 
                 auth_info = cloud.config["auth"]
+                logger.debug(f"OpenStack auth_url: {auth_info['auth_url']}")
 
                 if "username" in auth_info and "password" in auth_info:
+                    logger.debug(f"OpenStack username: {auth_info['username']}"),
                     return AuthPassword(
                         auth_url=auth_info["auth_url"],
                         username=auth_info["username"],
@@ -162,6 +164,7 @@ class OpenStackProvider(Provider):
                     "application_credential_id" in auth_info
                     and "application_credential_secret" in auth_info
                 ):
+                    logger.debug(f"OpenStack application_credential_id: {auth_info['application_credential_id']}")
                     return AuthPassword(
                         auth_url=auth_info["auth_url"],
                         application_credential_id=auth_info[
@@ -183,6 +186,7 @@ class OpenStackProvider(Provider):
                 OptionError,
                 MissingRequiredOptions,
             ) as terr:
+                logger.debug(f"Error during authentication: {terr}")
                 err_msg = (
                     "OpenStack credentials not provided or not properly configured. "
                     + "Authentication failed."
```

Such logging could be useful also when debugging issues in production.

src/mrack/providers/openstack.py
```diff
@@ -117,8 +121,102 @@ async def _openstack_gather_responses(self, *calls):
...
+        """
+        try:
+            # Create session from environment variables
+            return AuthPassword()
+        except TypeError:
+            # Fallback to create session from clouds.yaml file
+            try:
+                config = os_client_config.OpenStackConfig()
+
+                # Get the cloud specified in provisioning-config
```
pvoborni, line 143: Here is a question. Should `OS_CLOUD` override `profile` from provisioning config?  

The use case is to pick with what credentials to authenticate in LTE. As users might have different name or preferred cloud than the one in provisioning config. And current provisioning config wins. 

It is not that easy to pick the right cloud by changing the value provisioning config.

E.g. if we do in, e.g. transformer something like this:

```diff
diff --git a/src/mrack/transformers/openstack.py b/src/mrack/transformers/openstack.py
index 1f0f1e9..854ba01 100644
--- a/src/mrack/transformers/openstack.py
+++ b/src/mrack/transformers/openstack.py
@@ -14,6 +14,7 @@
 
 """OpenStack transformer module."""
 import logging
+import os
 
 from mrack.errors import ProvisioningConfigError
 from mrack.providers.provider import STRATEGY_ABORT
@@ -41,12 +42,17 @@ class OpenStackTransformer(Transformer):
     async def init_provider(self):
         """Initialize associate provider and transformer display name."""
         self.dsp_name = "OpenStack"
+
+        os_cloud = os.environ.get('OS_CLOUD')
+        if not os_cloud:
+            os_cloud = self.config.get("profile", DEFAULT_CLOUD_PROFILE)
+
         await self._provider.init(
             image_names=self.config["images"].values(),
             networks=self.config["networks"],
             strategy=self.config.get("strategy", STRATEGY_ABORT),
             max_retry=self.config.get("max_retry", DEFAULT_ATTEMPTS),
-            cloud_profile=self.config.get("profile", DEFAULT_CLOUD_PROFILE),
+            cloud_profile=os_cloud,
             keypair=self.config["keypair"],
             pubkey=self.config["pubkey"],
```

src/mrack/providers/openstack.py
dav-pascual: Logging messages added

requirements.txt
```diff
@@ -5,3 +5,5 @@ click>=7.0.0
 pyyaml>=5.0.0
 asyncopenstackclient>=0.8.1
 testcloud>=0.5.0
+aiofiles>=23.1.0
```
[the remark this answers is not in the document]
pvoborni: What about this comment?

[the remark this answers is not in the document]
dav-pascual: on it! :slightly_smiling_face:

pvoborni (approved):
LGTM

kaleemsiddiqu (approved):

PR #267, opened 2023-07-31
fix(Beaker): Exception has been thrown as raise missed argument

Tiboris:
Exception has been thrown as raise missed argument


dav-pascual (approved):
LGTM. Thanks for the patch @Tiboris :)

Tiboris:
David are these internal tests failing because of new way to authenticate ? Is it safe to merge?

dav-pascual:
> David are these internal tests failing because of new way to authenticate ? Is it safe to merge?

@Tiboris Correct, it's failing due to credentials. I will find a wy to fix them, in the meanwhile, i think we can merge this :)

Issue #269, opened 2023-09-19
8.0.8 semantic release does not support this command


Tiboris:
https://github.com/neoave/mrack/blob/7e0d0383d6c026e4ed671b3f1468d320f60a2eac/.github/workflows/test-and-release.yml#L62

Tiboris:
semantic-release version --print

dav-pascual:
@Tiboris thanks for reporting, will try it out :)

Tiboris:
and please remove the commit from master the command works i already tried

Tiboris:
maybe you can try to run release on your fork without the pypi token of course to see if everything except PYPI is fine

PR #270, opened 2023-09-19
fix: broken command print version in semantic release

dav-pascual:
From issue: https://github.com/neoave/mrack/issues/269


Tiboris (approved):
LGTM  thanks david please wait for master to heal :)

PR #271, opened 2023-09-19
chore: Bump python-semantic-release to v7.34.4

dav-pascual:
Version 7.33.1 of python-semantic-release action is failing in the container build. Updating action to this version will solve this issue.


Tiboris:
I will try this in my fork if release and build does the job

Tiboris:
this seems to help semantic release however pypi release should be done this way https://github.com/pypa/gh-action-pypi-publish i am merging release created sucessfully:

Tiboris (approved):
see above comment, adding ack

PR #275, opened 2023-10-05
chore(ci): Fix release workflow build step checking out wrong commit

dav-pascual:
actions/checkout checks out the latest commit at the time of the workflow trigger, 
hence any commit done during the workflow is not included. 
Fixing this to checkout the actual latest (release) commit.

### DIFF
--- a/.github/workflows/test-and-release.yml
+++ b/.github/workflows/test-and-release.yml
@@ -101,6 +101,10 @@ jobs:
     needs: [release]
     steps:
       - uses: actions/checkout@v4
+      - name: Checkout latest release commit
+        run: |
+          git fetch
+          git checkout $(git rev-parse FETCH_HEAD)
       - name: Set up Python
         uses: actions/setup-python@v4
         with:


dav-pascual:
Tested in https://github.com/dav-pascual/mrack/actions/runs/6417183083

Job checking out the actual latest release commit at https://github.com/dav-pascual/mrack/actions/runs/6417183083/job/17422433550#step:3:44

Commit done: https://github.com/dav-pascual/mrack/commit/c0fc82eec7878d45edbd95f4c219684f8747f19d

pvoborni:
I don't understand how does this fix #273.  There, the issue seems to be: 

`404 Client Error: Not Found for url: https://github.com/neoave/mrack/releases/download/v1.16.0/mrack-1.16.0.tar.gz` 

I.e. the release process did not upload the archive here: https://github.com/neoave/mrack/releases/tag/v1.16.0  

But how does the checkout fix it? I was looking at latest test&release run: https://github.com/neoave/mrack/actions/runs/6406506972 and that seems to fail on the fact the version was not bumped as the commit used "chore" type in commit (i think that "chore" doesn't bump version") it would need to be fix, feat or something else. 

The fact that the workflow runs with the version it was executed on seems good to me. Or what is the issue with it?

dav-pascual:
> I don't understand how does this fix #273. There, the issue seems to be:
> 
> `404 Client Error: Not Found for url: https://github.com/neoave/mrack/releases/download/v1.16.0/mrack-1.16.0.tar.gz`
> 
> I.e. the release process did not upload the archive here: https://github.com/neoave/mrack/releases/tag/v1.16.0
> 
> But how does the checkout fix it? I was looking at latest test&release run: https://github.com/neoave/mrack/actions/runs/6406506972 and that seems to fail on the fact the version was not bumped as the commit used "chore" type in commit (i think that "chore" doesn't bump version") it would need to be fix, feat or something else.
> 
> The fact that the workflow runs with the version it was executed on seems good to me. Or what is the issue with it?

@pvoborni Oops, didn't look at the packit logs and thought it was about the same issue. I'll unlink the issue.

The workflow run you looked at, indeed won't upload the new version to pypi because @Tiboris already did so manually.
In that workflow it also gets the right version (1.16.0) because the release commit was done in a previous workflow run, so the commit "is there" when the workflow starts: https://github.com/neoave/mrack/actions/runs/6406506972/job/17391355353#step:12:61

Look at this workflow run: https://github.com/neoave/mrack/actions/runs/6241434877
It's doing semantic release for 1.16 version, but it's trying to publish 1.15.1 because it didn't get the latest release commit done during the semantic release step.

Tiboris (approved):
Thanks @dav-pascual maybe in same step of building python dist packages we can use action to upload them to the release page as the files will be present at this point.

please note this will be affected for future releases and 1.16.0 is not proposed to fedora rpms yet.

PR #279, opened 2023-10-17
chore(ci): Temporarely remove packit tests

dav-pascual:
Until secret injection issue into test farm is solved


Tiboris (approved):
I support this :) <3

PR #280, opened 2023-11-01
fix: curate_auth_url func changed to non-async

kaleemsiddiqu:
await call for _curate_auth_url missing in session creation which caused regression.
function _curate_auth_url changed to non-async, tests updated to reflect the same


dav-pascual:
LGTM

PR #281, opened 2023-11-27
feat: Add async_timeout dependency to requirements.txt

dav-pascual:
This dependency was observed missing when installing mrack as a dependency in another python project


kaleemsiddiqu (approved):

dav-pascual:
Added dependency also to spec file

pvoborni (approved):

PR #286, opened 2024-02-01
feat: Provide beaker log links

skycastlelily:
Client like tmt may need to fetch the logs,store log contents for users


src/mrack/providers/beaker.py
```diff
@@ -277,16 +277,20 @@ def prov_result_to_host_data(self, prov_result, req):
     def _get_recipe_info(self, beaker_id, log_msg_start):
         """Get info about the recipe for beaker job id."""
         bkr_job_xml = self.hub.taskactions.to_xml(beaker_id).encode("utf8")
-
+        logs_dict = {}
         resources = []
         for recipe in eTree.fromstring(bkr_job_xml).iter("recipe"):
+            for logs in recipe.iter('logs'):
+                    for log in logs.iter('log'):
```
pvoborni, line 284: There is an indentation error: https://dev.azure.com/neoave/mrack/_build/results?buildId=1135&view=logs&j=e4354d11-29aa-59c2-ce04-0d9aced3ea05&t=37a28340-749f-5255-d3d8-b83a7b4239ca&l=16

pvoborni (commented):
Hi, 

thanks for the patch.  

Should I assume that similar contribution will be made to tmt to use it? In such case, I think it would make sense to also include a unit test which would test the thing tmt will expect. The reason is that mrack doesn't currently have any promise of internal API stability and thus some work can break this integration. Having a unit test might help highlight a potential breakage and thus either prevent it or start a conversation/change on tmt side.

skycastlelily:
Hi,
Thanks for your feedback.

>Should I assume that similar contribution will be made to tmt to use it?

Yes, and I'm gonna to write a unit test to cover this.

skycastlelily:
> Hi,
> 
> thanks for the patch.
> 
> Should I assume that similar contribution will be made to tmt to use it? In such case, I think it would make sense to also include a unit test which would test the thing tmt will expect. The reason is that mrack doesn't currently have any promise of internal API stability and thus some work can break this integration. Having a unit test might help highlight a potential breakage and thus either prevent it or start a conversation/change on tmt side.

Hi @pvoborni , I have scripted a unit test,which need the code from this https://github.com/neoave/mrack/pull/287,would you please review it ?Thanks:)

pvoborni (commented):
The code looks good to me. But the test fails in CI: https://dev.azure.com/neoave/mrack/_build/results?buildId=1141&view=logs&jobId=e4354d11-29aa-59c2-ce04-0d9aced3ea05&j=e4354d11-29aa-59c2-ce04-0d9aced3ea05&t=c546a923-41d2-519e-602b-c74183bf4a4c  - beaker lib tries to load `/etc/beaker/client.conf`

I amended the PR name to pass the check, but not sure why it failed again. Maybe it will be fixed in next push.

skycastlelily:
Sorry for the late response, I'm just back from my vacation.
I updated the code, the error should be fixed now:)

pvoborni (approved):
LGTM

PR #289, opened 2024-03-27
fix: prefix job_id with J: for _get_recipe_info

skycastlelily:
fix the tmt clean guest issue: https://github.com/teemtee/tmt/issues/2796

### DIFF
--- a/src/mrack/providers/beaker.py
+++ b/src/mrack/providers/beaker.py
@@ -390,6 +390,8 @@ class BeakerProvider(Provider):
         # the provider uses hostname from metadata of the VM which has failed
         # to validate the requirements for the provider
         log_msg_start = f"{self.dsp_name} [{host_name}]"
+        if host_id.isdigit():
+            host_id = "J:" + host_id
         if not host_id.startswith("J:"):
             logger.warning(
                 f"{log_msg_start} Job for host '{host_id}' does not exist yet"


skycastlelily:
The problem is:[delete_host]( https://github.com/neoave/mrack/blob/main/src/mrack/providers/beaker.py#L385),checks whether host_id.startswith("J:"),while tmt get the job_id from [_get_recipe_info](https://github.com/neoave/mrack/blob/main/src/mrack/providers/beaker.py#L277) ,and recipe['job_id'] doesn't has the J: prefix.

skycastlelily:
Hi,[pvoborni](https://github.com/pvoborni),would you please review this merge request,which is tiny and clear,and related to the merge request:https://github.com/neoave/mrack/pull/286:)

skycastlelily:
Sorry for the late reply,I'm just back from state holiday vacation.

The problem lies in the line of tmt:https://github.com/teemtee/tmt/blob/main/tmt/steps/provision/mrack.py#L696,create_server dose return a job_id with J:,but the def create function in tmt will return the info from _get_repcipe_info,ie,no J:, so  it will pass a job_id without J" when it calls mrack provider's delete_host,and the error happens.My fault,I thought create_server will return a job_id without J:,will be more careful next time when I submit a merge request.That being said,IMO we could and also should fix the problem from mrack side,as we never know whether other users will do the same thing:use the job_id from _get_recipe_info.So I update the merge request, would please review it,when you get some time to spare?Thanks:)

pvoborni (approved):
LGMT. 

In general, I don't like much the "magic" around the "J:", it would be better to have the beaker provider more refactored so the logic around the id is not on 3 different places and ideally tmt could avoid using internal method for getting current state.

skycastlelily:
Sorry for  the so late reply, I didn't realize I hadn't sent this out until
just now.

> it would be better to have the beaker provider more refactored so the
logic around the id is not on 3 different places and ideally tmt could
avoid using internal method for getting current state.

Yes,  it will be great if you could find some time  working on that:)

> Merged #289 <https://github.com/neoave/mrack/pull/289> into main.

Thanks:)

PR #290, opened 2024-04-19
fix(beaker): suppress 10_avc_check restraint plugin

pvoborni:
**fix(beaker): suppress 10_avc_check restraint plugin**

It has happpened that this plugin sometimes ran after the dummy task, it reported fail
that there was some AVC (probabaly from other thing) which then failed the job and thus
mrack treated this as a provisioning failure.

This patch instructs restraint to not run this plugin and thus avoid this situation.

**test: speed-up tests by mocking gethostbyaddr**

Tests that are using mrack outputs are slowed down by
socker.gethostbyaddr for adhoc IP addresses. This resolution mostly
fails and the test is slow (multiple seconds timeout for single IP).

With this, all python tests are executed within 1.5s.

**Notes**
The other patch  was created so that I can run the tests locally effectively.

### DIFF
--- a/src/mrack/providers/beaker.py
+++ b/src/mrack/providers/beaker.py
@@ -204,7 +204,11 @@ class BeakerProvider(Provider):
         recipe.addReservesys(duration=str(self.reserve_duration))
 
         for task in specs["tasks"]:
-            recipe.addTask(task=task["name"], role=task["role"])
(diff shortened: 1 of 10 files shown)


dav-pascual:
LGTM

dav-pascual (approved):

PR #293, opened 2024-05-30
chore: Update deprecated actions

dav-pascual:
Due to warnings shown in https://github.com/neoave/mrack/actions/runs/9302766806

It updates actions to new version with Node 20

### DIFF
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -10,8 +10,8 @@ jobs:
   pre-commit:
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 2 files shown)


dav-pascual:
Pre-commit checks shows it's working with no breaking changes

kaleemsiddiqu:
> Pre-commit check shows it's working with no breaking changes
Pre-commit checks only checkout/setup-python actions not upload/download artifacts but i think since these are just version upgrade of these actions so nothing will be impacted, adding my ack based on this justification

kaleemsiddiqu (approved):

PR #294, opened 2024-07-01
feat: Add support for translanting %pre and main body part of kickstart

skycastlelily:


skycastlelily:
This merge request is needed for adding support kickstart in tmt mrack plugin:)

pvoborni (commented):
I just quickly read the code, providing initial comment.

I'm thinking that it might be good to cover this functionality with more unit tests as now it only tests only part of the new scenario.

Mrack's documentation kinda sucks. But this feature is not very obvious and thus I'm thinking that some docs might help. Somewhere in https://github.com/neoave/mrack/tree/main/docs/guides

skycastlelily:
Ah,right, really sorry for making that kind of mistake, I should not only
pay attention to the part tmt use,I will keep that in mind,sorry again:)

skycastlelily:
Updated:)

src/mrack/transformers/beaker.py
```diff
@@ -86,17 +86,35 @@ def _get_ks_meta(self, host):
...
+                    res_ks_list += ["%pre"] + [res_ks_pre] + ["%end"]
+            if res_ks:
+                res_ks_list += [res_ks]
+            if res_ks_post:
+                if res_ks_post.startswith("%post"):
+                    res_ks_list += [res_ks_post]
+                else:
+                    res_ks_list += ["%post"] + [res_ks_post] + ["%end"]
+        else:
+            res_ks_list = ["%post"]
```
pvoborni, line 110: I think in this case, when ssh keys are defined, there will be 2 `%post` section. I don't know if it is correct. But might not be and thus highlighting it.

skycastlelily:
yes,that is desired output, beaker is able to translate more than one %post
section:)

skycastlelily:
>yes,that is desired output, beaker is able to translate more than one %post section:)

Hi @pvoborni ,please feel free to tell me if you have any other concerns:)

dav-pascual:
> > yes,that is desired output, beaker is able to translate more than one %post section:)
> 
> Hi @pvoborni ,please feel free to tell me if you have any other concerns:)

@skycastlelily He is OOO right now, so I will handle the review :)

skycastlelily:
Hi,thanks for your info and help in advance:)

dav-pascual:
Hi @skycastlelily! In principle, changes LGTM
But before approving, fix the test failures first please `FileNotFoundError: [Errno 2] No such file or directory: 'key_two'`. You will probably need to mock that part.

skycastlelily:
> You will probably need to mock that part.

you mean mock [_allow_ssh_keys](https://github.com/neoave/mrack/blob/main/src/mrack/transformers/beaker.py#L101 )  ?By that way we won't be able to check whether _allow_ssh_keys works well or not, I have updated the test case to resolve the issue you mentioned, please feel free to tell me if that way is not expected.

> You can run pytest in your local env installing the deps from test-requirements.txt

Thanks for your info,gonna to check:)

skycastlelily:
Updated again,should be good now:)Please note,the original doesn't work well,as it omits a comma after "\n".join(res_ks_append) ,as a result the job submit will have lots of <ks_append><![CDATA[one-letter]]></ks_append><ks_append> lines

dav-pascual:
> > You will probably need to mock that part.
> 
> you mean mock [_allow_ssh_keys](https://github.com/neoave/mrack/blob/main/src/mrack/transformers/beaker.py#L101) ?By that way we won't be able to check whether _allow_ssh_keys works well or not, I have updated the test case to resolve the issue you mentioned, please feel free to tell me if that way is not expected.
> 
> > You can run pytest in your local env installing the deps from test-requirements.txt
> 
> Thanks for your info,gonna to check:)

I had in mind something like a '@patch' mock style, but this works too.

> Updated again,should be good now:)Please note,the original doesn't work well,as it omits a comma after "\n".join(res_ks_append) ,as a result the job submit will have lots of <ks_append></ks_append><ks_append> lines

I'm not sure I have understood this lastest change. As far as I know, adding a comma to the list doesn't make any difference. What was wrong before that now it is not? And why is it necessary to sort the ssh keys?

skycastlelily:
>  I'm not sure I have understood this lastest change. As far as I know, adding a comma to the list doesn't make any 
> difference. What was wrong before that now it is not? 

The comma is there because [this line](https://github.com/neoave/mrack/blob/main/src/mrack/transformers/beaker.py#L96) will make the ks_append string to a long list whose elements containing only one letter. This mr is good without that comma because I have brackets the strings with [], or we will get a  job like [this one ](https://beaker.engineering.redhat.com/jobs/clone?job_id=9570492)

>And why is it necessary to sort the ssh keys?

To make the test pass:)Set is unordered, key_two's content may be printed before key_one's.

dav-pascual:
@skycastlelily Great, thanks for the changes. Let's merge!

dav-pascual (approved):

skycastlelily:
Thanks:)

PR #295, opened 2024-07-11
feat: update pytest-mh output to work with latest version

pbrezina:
pytest-mh implemented a generic connection interface so one
can choose between ssh, podman and docker. This introduced
a breaking change in the configuration.

The `ssh` field was changed to a generic `conn` field that
specifies the connection type.

More information can be found in the pytest-mh pull request.

Depends on:
* https://github.com/next-actions/pytest-mh/pull/67


kaleemsiddiqu (approved):
Tested successfully in downstream env, so adding ack

PR #302, opened 2024-11-06
refactor: Return empty list when there is no content in res_ks_list

skycastlelily:


dav-pascual:
Thanks for the PR. LGTM.
Did you make this change to fix some edge case (e.g. empty ks_append dict)? In that case it would a fix rather than a refactor

skycastlelily:
> Thanks for the PR. LGTM.

Thanks for your review.

> Did you make this change to fix some edge case?

Yep, this mr is for only metadata provided case.

>  In that case it would a fix rather than a refactor.

Gonna to update the commit message.

dav-pascual (approved):

PR #303, opened 2024-12-04
fix: Translate job-group properly

skycastlelily:


skycastlelily:
Block  tmt mr: https://github.com/teemtee/tmt/pull/3397 :)

skycastlelily:
Hi Dav, thanks for your review, I have updated the commit message, I guess we are good now, and sorry for failing to note that before I submit this mr :sweat_smile:

dav-pascual:
Hi @skycastlelily 
Thank you for your changes. In principle LGTM.
The reason behind this change is that previous key was wrong, right?

skycastlelily:
> The reason behind this change is that previous key was wrong, right?

Yes ,
BeakerJob use job-group key instead of group.
I have moved the above line here,hope we can get rid of the PR header format error:)

dav-pascual:
LGTM. Nevermind about the failing check, idk why it's still failing

PR #307, opened 2025-03-10
fix: Translate nested or/and constraint properly

skycastlelily:


skycastlelily:
This mr fixes the following  tmt issue:

If users specify a hardware with   nested and/or ,say:

    ```
hardware:
        or:
          - hostname: == dummy1.redhat.com
          - or:
              - hostname: == dummy2.redhat.com
              - or:
                  - hostname: == dummy3.redhat.com
                  - hostname: == dummy4.redhat.com

```

They will get a job  containing the following hostrequires 

```
      <hostRequires>
        <and>
          <or>
            <hostname op="==" value="dummy1.redhat.com"/>
            <or/>
          </or>
        </and>
      </hostRequires>
```

File: src/mrack/providers/beaker.py [hunk 1]
```diff
@@ -159,6 +159,29 @@ async def utilization(self):
         """Check percentage utilization of given provider."""
         return 0
 
+    def trans_constraint(self, host_requires, host_recipe):
```
dav-pascual, line 162: could you rename the method to `_translate_constraint`?

dav-pascual:
Would you have time to add a testcase for this in `tests/unit/test_beaker.py`?

skycastlelily:
Hi,
Thanks for your review.

> Please add this to the commit message

How about we open a mrack issue,and paste the link to the commit message?:)

dav-pascual:
> Hi, Thanks for your review.
> 
> > Please add this to the commit message
> 
> How about we open a mrack issue,and paste the link to the commit message?:)

That would work too :)

dav-pascual:
@skycastlelily Please, check out the pipelines. There are some flake8 errors

tests/unit/test_beaker.py
```diff
@@ -91,3 +92,72 @@ async def test_beaker_job_creation(self, mock_beaker_conf):
         # so that provisioning won't fail on AVC denials
         xml = job.toxml()
         assert '<param name="RSTRNT_DISABLED" value="10_avc_check"/>' in xml
+
+    @pytest.mark.asyncio
+    async def test_translate_constraint(self, mock_beaker_conf):
+        provider = BeakerProvider()
+        await provider.init(self.distros, self.timeout, self.reserve_duration)
+        req_node = xml_doc().createElement("and")
+        host_requires = {
```
dav-pascual, line 101: In principle, it looks good. But I would change the following:

- Instead of asserting a pretty XML, it should be normalized, and them compared. We could use e.g. minidom. Check an example here https://www.geeksforgeeks.org/compare-two-xml-files-in-python/
- Take host_requires out of the method so it can be reused if needed in future tests. Either as a class variable or a function that returns the object (see other test files)

dav-pascual:
Skipping unit tests for now due to time constraints. This was tested manually in integration with tmt

PR #329, opened 2026-03-26
fix: Pin setuptools<82 to restore pkg_resources for beaker-client

mrizwan93:
setuptools 82.0.0 removed pkg_resources, which beaker-client imports at module level. This broke test collection for beaker-related tests.


dav-pascual (approved):
LGTM

PR #330, opened 2026-04-10
feat: add SSM Parameter Store support for AWS AMI resolution

dav-pascual:
Allow resolving the latest AMI for a specific OS version (e.g. Windows Server 2025) without hardcoding an AMI ID, by using AWS SSM Parameter Store paths in provisioning-config.yaml.

Assisted-by: Claude (Anthropic)

### DIFF
--- a/src/mrack/providers/aws.py
+++ b/src/mrack/providers/aws.py
@@ -61,6 +61,13 @@ class AWSProvider(Provider):
         """Get provider name."""
         return self._name
 
+    @property
+    def ssm_client(self):
+        """Get SSM client, creating it on first use."""
+        if self._ssm_client is None:
+            self._ssm_client = boto3.client("ssm")
+        return self._ssm_client
+
     async def init(
         self,
         ssh_key,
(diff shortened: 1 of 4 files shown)


src/mrack/providers/aws.py
```diff
@@ -174,6 +203,14 @@ def load_image(self, req):
             name = tag_def["name"]
             filters.append({"Name": f"tag:{name}", "Values": [tag_def["value"]]})
 
+        # by SSM parameter
+        elif isinstance(image_def, dict) and "ssm" in image_def:
+            ssm_path = image_def["ssm"]
+            response = self.ssm_client.get_parameter(Name=ssm_path)
```
mrizwan93, line 209: would be good to check if `ami_id` exist in `self.ssm_resolved` first before calling get_parameter() ?

dav-pascual: get_image() is already handling that. Same way as checking self.amis cache, it's the method in charge of doing these checks

mrizwan93 (approved):

PR #331, opened 2026-06-10
feat: configure root EBS volume size via disksize(aws)

mrizwan93:
EC2 instances previously inherited only the AMI snapshot size for the root volume, which is often too small for CI prep (e.g. client hosts running out of disk during Ansible runs).
Resolve disksize from host metadata, aws.groups, or aws.options and pass VolumeSize in BlockDeviceMappings when launching instances. Document defaults in the sample provisioning config and add unit tests.

### DIFF
--- a/src/mrack/data/provisioning-config.yaml
+++ b/src/mrack/data/provisioning-config.yaml
@@ -63,9 +63,10 @@ aws:  # aws provider config
         win-2022-latest:  # find latest AMI via AWS SSM parameter
             ssm: /aws/service/ami-windows-latest/Windows_Server-2022-English-Full-Base
 
-    flavors:
-        # list of available flavours to ask from provider  for vm specs
-        default: t2.nano  # e.g. this is 1vcpus 0.5GB ram
+    # Deprecated: prefer aws.groups for flavor/disksize (groups take precedence).
+    # aws.flavors remains supported for backwards compatibility.
+    # flavors:
+    #     default: t2.nano  # e.g. this is 1vcpus 0.5GB ram
 
     keypair: mrack  # this name points to uploaded and available in aws ec2 key configuration
     security_group: sg-0ad1bab3a850e70fc  # setup the system firewall using one of existing sec groups
@@ -84,6 +85,17 @@ aws:  # aws provider config
         mrack: "True"
         persistent: "False"
     delete_volume_on_termination: True # instance volume is deleted on termination, default: True
+    # Per-group instance type and root volume (takes precedence over options)
+    options:
+        flavor: t2.nano  # default instance type
+        disksize: 10  # default root EBS volume size in GiB
+    groups:
+        client:
+            flavor: t3.micro
+            disksize: 20
+        ipaserver:
+            flavor: t3.medium
+            disksize: 40
     spot: True  # request spot EC2 instances, default: False
     resolve_host: False  # resolve hostname from IP on output generation, default: True
 
--- a/src/mrack/providers/aws.py
+++ b/src/mrack/providers/aws.py
@@ -374,6 +374,7 @@ class AWSProvider(Provider):
         specs = deepcopy(req)  # work with own copy, do not modify the input
 
         del_vol = specs.get("delete_volume_on_termination", True)
+        disksize = specs.get("disksize")
 
         name = req.get("name")
         # creating unique name for instance (visible in aws ec2 WebUI)
@@ -394,6 +395,13 @@ class AWSProvider(Provider):
 
         logger.debug(f"{log_msg_start} Tagging instance with: {object2json(taglist)}")
 
+        ebs = {"DeleteOnTermination": del_vol}
+        if disksize is not None:
+            ebs["VolumeSize"] = disksize
+            logger.info(
+                f"{log_msg_start} Root EBS volume size: {ebs['VolumeSize']} GiB"
+            )
+
         request = {
             "ImageId": self.get_image(specs).image_id,
             "MinCount": 1,
@@ -404,9 +412,7 @@ class AWSProvider(Provider):
             "BlockDeviceMappings": [
                 {
                     "DeviceName": "/dev/sda1",
-                    "Ebs": {
-                        "DeleteOnTermination": del_vol,
-                    },
+                    "Ebs": ebs,
                 },
             ],
             "TagSpecifications": [{"ResourceType": "instance", "Tags": taglist}],
--- a/src/mrack/transformers/aws.py
+++ b/src/mrack/transformers/aws.py
@@ -25,7 +25,6 @@ class AWSTransformer(Transformer):
 
     _config_key = CONFIG_KEY
     _required_config_attrs = [
-        "flavors",
         "images",
         "credentials_file",
         "keypair",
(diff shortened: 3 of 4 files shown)


src/mrack/data/provisioning-config.yaml
```diff
@@ -84,6 +84,13 @@ aws:  # aws provider config
         mrack: "True"
         persistent: "False"
     delete_volume_on_termination: True # instance volume is deleted on termination, default: True
+    options:
+        disksize: 10  # default root EBS volume size in GiB
+    groups:
+        client:
+            disksize: 20
+        ipaserver:
+            disksize: 40
```
dav-pascual, line 98: The problem I see with this approach is that we are duplicating the flavor, which can bring inconsistencies. In this same provisioning-config.yaml example, you'd define `client` and `ipaserver` in groups, but not in flavor, which would make the mrack fallback to the default one.

We could consolidate on groups e.g.
```
  aws:
    flavors:              # old, still works
      ipaserver: t3.large
    groups:               # new, takes precedence
      ipaserver:
        flavor: t3.xlarge
        disksize: 40
```

while still supporting old flat level flavors to not introduce breaking changes

The change would be simple, in `_get_flavor()`,  you'd add something like this before the existing logic:

```
  groups = self.config.get("groups", {})
  group_flavor = groups.get(host.get("size") or host["group"], {}).get("flavor")
  if group_flavor:
      return group_flavor
```

or better we can generalize the lookup:
```
 def _get_group_value(self, host, attr):
      """Look up attr from groups config for this host's group."""
      groups = self.config.get("groups", {})
      return groups.get(host["group"], {}).get(attr)
```

libvirt provider does something similar

WDYT

src/mrack/providers/aws.py
```diff
@@ -404,9 +412,7 @@ async def create_server(self, req):
             "BlockDeviceMappings": [
                 {
                     "DeviceName": "/dev/sda1",
```
dav-pascual, line 414: Not part of this PR, but some AMI might define different device name. Worth checking

src/mrack/providers/aws.py
```diff
@@ -394,6 +395,13 @@ async def create_server(self, req):
 
         logger.debug(f"{log_msg_start} Tagging instance with: {object2json(taglist)}")
 
+        ebs = {"DeleteOnTermination": del_vol}
+        if disksize is not None:
+            ebs["VolumeSize"] = int(disksize)
```
dav-pascual: harmless, but int cast is applied twice, here and in the transformer. It should be a job of the transformer

tests/unit/test_aws_transformer.py
```diff
@@ -198,6 +198,43 @@ async def test_user_data_for_multiple_windows_os(self):
...
+        config = _aws_provisioning_config()
+        config["aws"]["groups"] = {"ipaclient": {"disksize": 20}}
+        providers.register(AWS, AWSProvider)
+        transformer = MockedAWSTransformer()
+        host = _host("client", "client", "ipaclient", "rhel-8.5")
+        host["disksize"] = 30
+        metadata = {"domains": [{"name": DOMAIN, "type": "mixed", "hosts": [host]}]}
+        await transformer.init(config, metadata)
+        req = transformer.create_host_requirement(host)
+        assert req["disksize"] == 30
```
dav-pascual, line 251: It would be worth a check also for default `options.disksize`

src/mrack/providers/aws.py
mrizwan93: opened https://github.com/neoave/mrack/issues/332

src/mrack/transformers/aws.py
```diff
@@ -14,10 +14,27 @@
...
+        "xsmall",
+        "small",
+        "medium",
+        "large",
+        "xlarge",
+        "xxlarge",
+        "xxxlarge",
+        "disk-large",
+    }
+)
```
dav-pascual: mrack is designed as an agnostic provider, and should not contain hardcoded flavors

src/mrack/data/provisioning-config.yaml
```diff
@@ -64,7 +64,7 @@ aws:  # aws provider config
             ssm: /aws/service/ami-windows-latest/Windows_Server-2022-English-Full-Base
 
     flavors:
-        # list of available flavours to ask from provider  for vm specs
+        # list of available flavours to ask from provider for vm specs
         default: t2.nano  # e.g. this is 1vcpus 0.5GB ram
```
dav-pascual: I would comment out flavors specifying is deprecated in favor of groups, though it's still supported, `groups` taking precedence

src/mrack/transformers/aws.py
```diff
@@ -113,5 +190,9 @@ def create_host_requirement(self, host):
             "user_data": self._find_value(host, "user_data", "user_data", host["os"]),
         }
 
+        disksize = self._get_disksize(host)
+        if disksize is not None:
+            req["disksize"] = disksize
+
```
dav-pascual, line 133: I think the implementation of AWS transformer got a bit complex with the last refactor when it should be simpler. Check out the implementation for Virt wich also uses groups: https://github.com/neoave/mrack/blob/main/src/mrack/transformers/virt.py#L62 

From a simple _get_host_option we get the params and add them to the req. The only difference would be adding the retrocompatibility support for plain flavors. It could be something like:

```
  def _get_host_option(self, host, name):
      # standard 3-level: host → groups.<group> → options
      default_options = self.config.get("options", {})
      group_options = self.config.get("groups", {}).get(host["group"], {})
      val = host.get(name) or group_options.get(name) or default_options.get(name)

      # backwards compat: fall back to legacy aws.flavors map
      if val is None and name == "flavor":
          val = super()._get_flavor(host)

      return val
```

dav-pascual (changes requested):

src/mrack/data/provisioning-config.yaml
dav-pascual: I would leave this commented out and move the default to options

src/mrack/transformers/aws.py
dav-pascual: Thanks for the changes. One last comment. There's no need to check None for `disksize`, that is handled by the AWS provider module. Then to simplify, we can have in the method:

```
for option in [                                                                                                                                                                                           
    "flavor",                                                                                                                                                                                             
    "disksize",                                                                                                                                                                                           
]:                                                                                                                                                                                                        
    req[option] = self._get_host_option(host, option) 
```

tests/unit/test_aws_transformer.py
```diff
@@ -198,6 +198,116 @@ async def test_user_data_for_multiple_windows_os(self):
...
+        metadata = {"domains": [{"name": DOMAIN, "type": "mixed", "hosts": hosts}]}
+        await transformer.init(config, metadata)
+
+        master_req = transformer.create_host_requirement(hosts[0])
+        runner_req = transformer.create_host_requirement(hosts[1])
+        assert master_req["flavor"] == "t3.medium"
+        assert master_req["disksize"] == 40
+        assert runner_req["flavor"] == "t3.small"
+        assert runner_req["disksize"] == 25
+
```
dav-pascual: Great. We can probably drop `test_flavor_from_groups_takes_precedence` and `test_flavor_from_groups_takes_precedence` as they do redundant checks covered by the other

dav-pascual (changes requested):

tests/unit/test_aws_transformer.py
dav-pascual: My bad i repeated the same test, the other was `test_ipa_style_uses_inventory_group_from_metadata`

dav-pascual (approved):
LGTM!

PR #333, opened 2026-09-22
fix(aws): give size precedence over group for flavor/disksize

dav-pascual:
## Summary
AWS picked flavor/disksize from host metadata, then group, then options -- size was never checked, unlike OpenStack where size already wins over group. Now size ranks between metadata and group: a host with group: ipaserver and size: xxlarge gets xxlarge's flavor/disksize instead of ipaserver's. A size entry only overrides the fields it actually sets, so it still falls back to group for anything it leaves out.

## Test plan
- [x] `pytest tests/unit/test_aws_transformer.py` -- 15/15 passed
- [x] Manual check: group: ipaserver + size: xxlarge now resolves from xxlarge instead of getting stuck on ipaserver


rvala02 (approved):
LGTM

mrizwan93 (approved):

